PR #7, opened 2022-01-04
Bugfix/JSSNMSHDD-2746 check for creation change status in auto accept relationship creation changes module

jkoenig134:


src/modules/autoAcceptRelationshipCreationChanges/AutoAcceptRelationshipCreationChangesModule.ts
```diff
@@ -37,8 +36,13 @@ export default class AutoAcceptRelationshipCreationChangesModule extends Connect
         });
     }
 
-    private isIncomingRelationshipCreationChange(event: RelationshipChangedEvent) {
-        return event.data.changes.length === 1 && event.data.changes[0].request.createdBy !== this.currentIdentity;
+    private isIncomingPendingRelationshipCreationChange(event: RelationshipChangedEvent) {
+        const data = event.data;
+
+        if (data.changes.length === 1) return false;
```
stnmtz, line 42: I don't think that is correct here, this will always return false for any incoming change ;)

stnmtz (changes requested):

stnmtz (approved):

PR #8, opened 2022-01-10
Add tini init and run as node user

CSNWEB:
Both is recommmended by the nodejs docker working group:
https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md

- Tini makes sure that kernel signals are handled properly, otherwise nodejs containers can take a while to shutdown sometimes.
- It is also recommended to run the container as non-root node user

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches-ignore:
       - main
+  pull_request:
+    types: [opened, edited]
 
 jobs:
   test:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 2.1.4
+
+-   run Connector with `node` user instead of `root`
+-   run `tini` together with the Connector to ensure kernel events will be propagated to the Connector
+
 ## 2.1.3
 
 -> SDK 1.1.6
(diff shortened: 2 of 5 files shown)


tnotheis (changes requested):
I created a pull request with some minor changes that merges into your branch. As soon as you merged it we will merge this one here as well.

tnotheis:
@CSNWEB We noticed that a container built with your proposed Dockerfile doesn't start because the `node` user lacks write permission on the `/usr/app/logs` folder. 
We tried to find a quick solution for it it but the only thing we came up with was (under certain circumstances) a breaking change for people that mount their log files to the host machine. We will try to find a solution for this as soon as we have the capacities.

I guess for now it would be best to remove the line where you set the user so that we can merge this PR.

tnotheis:
Another problem with the default docker logs is that after you recreate the container (e.g. in case of a version upgrade) all the logs created in the past will be lost. Whereas if you mount the logs to the host machine, they will be persisted across container recreations.

Nonetheless I am not fully convinced of the current logging solution either. I guess it would be best to have some external service we can send the logs to. But that's not on top of our priority list at the moment.

jkoenig134:
Hi @CSNWEB! As TINI was added a while ago we finally switched from the `root` user to the `node` user.

This will land with the `3.0.0` release of the Connector that will drop soon and can already be tested using the `beta` tag. Make sure to read the updated information about [log file mounting](https://enmeshed.eu/versions/2.0.0/integrate/connector-installation#log-file-mounting) when setting up the connector!

PR #120, opened 2023-11-17
Feature/pubsub publisher

jkoenig134:


src/modules/pubsubPublisher/PubsubPublisherModule.ts
```diff
@@ -0,0 +1,54 @@
...
+import { DataEvent as RuntimeDataEvent } from "@nmshd/runtime";
+import { ConnectorRuntimeModule, ConnectorRuntimeModuleConfiguration } from "../../ConnectorRuntimeModule";
+
+export interface PubsubPublisherModuleConfiguration extends ConnectorRuntimeModuleConfiguration {
+    projectId: string;
+    topicName: string;
+    keyFile: string;
+}
+
+export default class PubsubPublisherModule extends ConnectorRuntimeModule<PubsubPublisherModuleConfiguration> {
```
stnmtz, line 12: We should use the official PubSub naming (uppercase S)

src/modules/pubsubPublisher/PubsubPublisherModule.ts
```diff
@@ -0,0 +1,54 @@
...
+    private pubsub: PubSub;
+    private topic: Topic;
+
+    public async init(): Promise<void> {
+        if (!this.configuration.projectId) throw new Error("Cannot start the module, the projectId is not defined.");
+        if (!this.configuration.topicName) throw new Error("Cannot start the module, the topic is not defined.");
+        if (!this.configuration.keyFile) throw new Error("Cannot start the module, the keyFile is not defined.");
+
+        this.pubsub = new PubSub({
```
stnmtz, line 21: We should add a logging output here, that a PubSub configuration for projectId and topicName is used.

src/modules/pubsubPublisher/PubsubPublisherModule.ts
```diff
@@ -0,0 +1,54 @@
...
+    private async handleEvent(event: Event) {
+        const data = event instanceof DataEvent || event instanceof RuntimeDataEvent ? event.data : {};
+        const buffer = Buffer.from(JSON.stringify(data));
+
+        await this.topic
+            .publishMessage({
+                attributes: { namespace: event.namespace },
+                data: buffer
+            })
+            .catch((e) => this.logger.error("Could not publish message", e));
```
stnmtz, line 46: Better debug output?
"Could not publish message to namespace xyz of topic abc"

src/modules/pubsubPublisher/PubsubPublisherModule.ts
```diff
@@ -0,0 +1,54 @@
...
+        this.pubsub = new PubSub({
+            projectId: this.configuration.projectId,
+            keyFile: this.configuration.keyFile
+        });
+
+        this.topic = this.pubsub.topic(this.configuration.topicName);
+        this.logger.info("Checking if topic exists...");
+
+        const topicExists = (await this.topic.exists())[0];
+        if (!topicExists) throw new Error(`Topic ${this.configuration.topicName} does not exist.`);
```
stnmtz, line 30: Shouldn't we add the projectId here?

src/modules/pubsubPublisher/PubsubPublisherModule.ts
stnmtz: What about errors thrown by PubSub (e.g. wrong keyfile?) - shouldn't we rethrow this with a logger output?

stnmtz (changes requested):
Please check official PubSub naming vs. Pubsub

With this, we have multiple Eventing modules:
- What happens if multiple modules are enabled in parallel?
- We might need module-based log outputs

src/modules/pubsubPublisher/PubsubPublisherModule.ts
jkoenig134: this will fail at the topic exists call and the module fails to initialize

src/modules/pubSubPublisher/PubSubPublisherModule.ts
```diff
@@ -0,0 +1,53 @@
...
+        this.pubSub = new PubSub({
+            projectId: this.configuration.projectId,
+            keyFile: this.configuration.keyFile
+        });
+
+        this.topic = this.pubSub.topic(this.configuration.topicName);
+        this.logger.info("Checking if topic exists...");
+
+        const topicExists = (await this.topic.exists())[0];
+        if (!topicExists) throw new Error(`Topic ${this.configuration.topicName} does not exist in the project '${this.configuration.projectId}'.`);
```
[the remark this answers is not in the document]
jkoenig134: it doesn't. The call to `this.topic.exists` can be rejected on internal errors -> line 32 will not be executed.

PR #124, opened 2023-12-05
Feature/attribute succession

J-Gann:


package.json
```diff
@@ -52,7 +52,7 @@
         "@js-soft/node-logger": "1.1.1",
         "@js-soft/ts-utils": "^2.3.3",
         "@nmshd/iql": "^1.0.0",
-        "@nmshd/runtime": "2.10.0",
+        "@nmshd/runtime": "3.0.0",
```
slavistan, line 55: 3.0.1 is out

package.json
```diff
@@ -24,7 +24,7 @@
         "tsc": "tsc",
         "tsc:watch": "tsc -w",
         "test:ci": "npm run build && jest -i",
-        "test:local": "npm run build && . .dev/scripts/testSetup.sh && jest",
+        "test:local": "npm run build --workspaces && . .dev/scripts/testSetup.sh && jest",
```
slavistan, line 27: What does the `--workspaces` flag do?

packages/sdk/src/endpoints/AttributesEndpoint.ts
```diff
@@ -2,19 +2,39 @@ import {
...
 import { Endpoint } from "./Endpoint";
 
 export class AttributesEndpoint extends Endpoint {
-    public async createAttribute(request: CreateAttributeRequest): Promise<ConnectorResponse<ConnectorAttribute>> {
-        return await this.post("/api/v2/Attributes", request);
+    public async createIdentityAttribute(request: CreateIdentityAttributeRequest): Promise<ConnectorResponse<ConnectorAttribute>> {
+        return await this.post("/api/v2/Attributes/IdentityAttribute", request);
+    }
+
+    public async createRelationshipAttribute(request: CreateRelationshipAttributeRequest): Promise<ConnectorResponse<ConnectorAttribute>> {
```
slavistan, line 24: Is there a reason this isnt' called `createAndShareRelationshipAttribute`. All of the new connector routes are named according to their use cases, except for this one. Obviously, this isn't a hard requirement, I'm just curious.

stnmtz, line 24: Is the Return Value for createRelationshipAttribute really ConnectorAttribute?

packages/sdk/src/endpoints/AttributesEndpoint.ts
```diff
@@ -2,19 +2,39 @@ import {
...
+        return await this.post("/api/v2/Attributes/RelationshipAttribute", request);
+    }
+
+    public async succeedIdentityAttribute(request: SucceedIdentityAttributeRequest): Promise<ConnectorResponse<ConnectorAttribute>> {
+        return await this.post(`/api/v2/Attributes/succeedIdentityAttribute/${request.predecessorId}`, request.successorContent);
```
stnmtz, line 29: AFAIK, these are the first routes starting with lowercase characters.

packages/sdk/src/types/attributes/requests/CreateRelationshipAttributeRequest.ts
```diff
@@ -0,0 +1,19 @@
+import { ConnectorAttributeValue } from "../ConnectorAttribute";
+
+export interface CreateRelationshipAttributeRequest {
```
stnmtz, line 3: Isn't it CreateAndShareRelationshipAttributeRequest?

packages/sdk/src/types/attributes/requests/CreateRelationshipAttributeRequest.ts
```diff
@@ -0,0 +1,19 @@
...
+    content: {
+        value: ConnectorAttributeValue;
```
stnmtz, line 5: Don't we have a ConnectorIdentityAttributeValue?

packages/sdk/src/types/attributes/requests/SucceedRelationshipAttributeRequest.ts
```diff
@@ -0,0 +1,10 @@
+import { ConnectorAttributeValue } from "../ConnectorAttribute";
+
+export interface SucceedRelationshipAttributeRequest {
+    predecessorId: string;
+    successorContent: {
+        value: ConnectorAttributeValue;
```
stnmtz, line 6: Don't we have a ConnectorRelationshipAttributeValue?

stnmtz (commented):
We've still some backwards-incompatibility questions ... :(

src/modules/coreHttpApi/controllers/AttributesController.ts
```diff
@@ -14,12 +14,42 @@ export class AttributesController extends BaseController {
     }
 
     @POST
+    @Path("/IdentityAttribute")
     @Accept("application/json")
-    public async createAttribute(request: any): Promise<Return.NewResource<Envelope>> {
```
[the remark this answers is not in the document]
J-Gann: Creation of Relationship and Identity Attributes are different in request parameters as well as in return type. Maybe use createAttribute for Identity Attributes and add a route "POST Attributes/RelationshipAttribute"?

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.7.0
+
```
jkoenig134, line 4: missing matching sdk version (after sdk major version bump)

packages/sdk/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@nmshd/connector-sdk",
-    "version": "2.2.4",
+    "version": "2.3.0",
```
jkoenig134, line 3: 3.0.0 (breaking change)

packages/sdk/src/types/attributes/requests/ShareAttributeRequest.ts
```diff
@@ -0,0 +1,4 @@
+import { NotifyPeerAboutIdentityAttributeSuccessionRequest } from "./NotifyPeerAboutIdentityAttributeSuccessionRequest";
+import { ShareIdentityAttributeRequest } from "./ShareIdentityAttributeRequest";
+
+export type ShareAttributeRequest = (ShareIdentityAttributeRequest & { "@type": "Share" }) | (NotifyPeerAboutIdentityAttributeSuccessionRequest & { "@type": "Notify" });
```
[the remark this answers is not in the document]
stnmtz: I guess this can be deleted?

packages/sdk/src/types/attributes/IQLQuery.ts
```diff
@@ -1,5 +1,5 @@
 export interface IQLQuery {
-    "@type"?: "IQLQuery";
+    "@type": "IQLQuery";
```
stnmtz, line 2: This is a breaking change where I don't see the reason. Could this be a bug within the Runtime?

test/attributes.test.ts
```diff
@@ -1,27 +1,29 @@
...
 beforeAll(async () => {
-    [client1] = await launcher.launch(1);
+    [client1, client2] = await launcher.launch(2);
+    await establishRelationship(client1, client2);
     client1Address = (await client1.account.getIdentityInfo()).result.address;
+    client2Address = (await client2.account.getIdentityInfo()).result.address;
 }, 30000);
 afterAll(() => launcher.stop());
 
 describe("Attributes", () => {
```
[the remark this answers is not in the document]
stnmtz: Nevermind, we were confused too - due to the fact we said that within the SDK it should no longer be used, the actual APIs of the Connector can't be tested with the SDK.

test/attributes.test.ts
```diff
@@ -166,8 +165,187 @@ describe("Execute AttributeQueries", () => {
                 }
             }
         });
+
         expect(executeRelationshipAttributeQueryResult).toBeSuccessful(ValidationSchema.ConnectorAttribute);
 
         expect(executeRelationshipAttributeQueryResult.result.content.value.value).toBe("AString");
     });
+
+    // TODO: Tests lead to timout errors in the syncUntilHasMessages method => remove tests?
```
stnmtz, line 174: Could it be, that another sync (another worker?) called the /sync route and stole the message?

test/iqlquery.test.ts
```diff
@@ -75,6 +75,7 @@ test("Local IQL Query", async () => {
     for (const e of table) {
         const queryRequest: ExecuteIQLQueryRequest = {
             query: {
+                "@type": "IQLQuery",
```
stnmtz, line 78: Could this be a bug? Why do we need to add the IQLQuery as type here? (See type definition within the SDK above). From my point of view, it is not necessary within the ExecuteIQLQueryRequest to require a @ type definition?

packages/sdk/src/endpoints/AttributesEndpoint.ts
```diff
@@ -1,20 +1,52 @@
...
-        return await this.post("/api/v2/Attributes", request);
+    public async createIdentityAttribute(request: CreateIdentityAttributeRequest): Promise<ConnectorResponse<ConnectorAttribute>> {
+        return await this.post("/api/v2/Attributes/CreateIdentityAttribute", request);
+    }
+
+    public async createAndShareRelationshipAttribute(request: CreateAndShareRelationshipAttributeRequest): Promise<ConnectorResponse<ConnectorRequest>> {
+        return await this.post("/api/v2/Attributes/CreateAndShareRelationshipAttribute", request);
+    }
+
+    public async succeedAttribute(request: SucceedAttributeRequest): Promise<ConnectorResponse<SucceedAttributeResponse>> {
```
stnmtz, line 32: I'm confused, we now mix two different options: I find it strange to add a new "convenience route" succeedAttribute, but remove the old route createAttribute in the SDK (but still allow it as a route in the Connector to be backwards-compatible).

I see two options:
a) We have no convenience routes in the SDK, but POST /Attributes stays as deprecated in the Connector. Then succeedAttribute should be removed from the SDK and from the Connector. But we loose the ability to test /Attributes.
b) We have convenience routes within the Connector and its SDK, both routes won't be deprecated, then createAttribute needs to stay in the SDK. Thus we do not need a major version inc and still be backwards compatible (and can test all Connector routes)

@jkoenig134 @slavistan Could we have a call tomorrow to clarify this?

I would vote for b).

packages/sdk/src/types/attributes/requests/CreateRelationshipAttributeRequest.ts
slavistan: There exists `ConnectorRelationshipAttribute`, which seems to fit here.

packages/sdk/src/types/attributes/requests/SucceedIdentityAttributeRequest.ts
```diff
@@ -0,0 +1,11 @@
+import { ConnectorAttributeValue } from "../ConnectorAttribute";
+
+export interface SucceedIdentityAttributeRequest {
+    predecessorId: string;
+    successorContent: {
+        value: ConnectorAttributeValue;
```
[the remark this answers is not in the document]
slavistan: There's a `ConnectorIdentityAttribute`, which I'll use here.

src/modules/coreHttpApi/controllers/AttributesController.ts
```diff
@@ -16,10 +16,77 @@ export class AttributesController extends BaseController {
     @POST
     @Accept("application/json")
     public async createAttribute(request: any): Promise<Return.NewResource<Envelope>> {
```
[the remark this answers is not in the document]
slavistan: What do you mean exactly by _try_? I can replace `any` with `unknown`, which the forces me to manually cast the request to some non-`unknown` type.

What's the problem with `any` here?

packages/sdk/src/endpoints/AttributesEndpoint.ts
slavistan: I agree, there's an inconsistency. I think it's better to present a unified interface to integrators, so the connector API and the SDK should match.

A third option is to not implement the `succeedAttribute` convenience route altogether. It's new and it was only created under the false assumption, that it already existed in the past.

src/modules/coreHttpApi/controllers/AttributesController.ts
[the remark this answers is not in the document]
jkoenig134: the problem is that you missed a null pointer bug b/c of any. `any` is fine if you don't do anything other than just passing it to the runtime, but can EXPLODE if you handle it like you did it.

src/modules/coreHttpApi/controllers/AttributesController.ts
```diff
@@ -16,10 +16,76 @@ export class AttributesController extends BaseController {
...
-        const selfAddress = (await this.transportServices.account.getIdentityInfo()).value.address;
-        if (request?.content?.owner !== selfAddress) throw new Error("You are not allowed to create an attribute that is not owned by yourself");
+        /* We left 'owner' and '@type' optional in the openapi spec for
+         * backwards compatibility. If set, they have to be removed here or the runtime
+         * use case will throw an error. */
+        if (typeof request.content?.owner !== "undefined") delete request.content.owner;
+        if (request.content?.["@type"] === "IdentityAttribute") delete request.content["@type"];
```
jkoenig134, line 23:
```suggestion
        if (typeof request?.content?.owner !== "undefined") delete request.content.owner;
        if (request?.content?.["@type"] === "IdentityAttribute") delete request.content["@type"];
```

packages/sdk/package.json
jkoenig134: @slavistan you resolved but not changed this.

jkoenig134: please don't forget to change it in the changelog too

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+-> SDK 2.3.0
+
+-   fix: POST /api/v2/Attributes will no longer accept RelationshipAttributes, as this was unintentional and led to wrong behavior
+-   add usecases for attribute succession
```
jkoenig134, line 8:
```suggestion
-   add routes for attribute succession
```

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+-   refactor existing attribute use cases
```
jkoenig134, line 9: the user will never see a refactoring - this is no information for the user and should be removed or better specified.

jkoenig134: (and we have no "use cases" in the connector)

packages/sdk/src/types/attributes/requests/ShareAttributeRequest.ts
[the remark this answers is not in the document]
jkoenig134: and `@type` Notify is strange as well

packages/sdk/src/types/attributes/requests/SucceedIdentityAttributeRequest.ts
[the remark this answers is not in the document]
slavistan: Just kidding, the `ConnectorIdentityAttribute` requires the `@type` field, which we don't need here. Same thing for the `ConnectorRelationshipAttribute` with regards any routes relating to relationship attributes. I'll go back to using a custom interface.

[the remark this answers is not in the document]
jkoenig134: wtb using `Omit` on `@type`?

packages/sdk/package.json
slavistan: Oh, my bad. I misread `3.0.0` as `2.3.0`.

src/modules/coreHttpApi/controllers/AttributesController.ts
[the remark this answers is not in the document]
jkoenig134: don't get me wrong, we use any everywhere - and at some places there could be bugs because of this.. I think thats all Thomas wants to say :D

packages/sdk/src/types/attributes/requests/SucceedIdentityAttributeRequest.ts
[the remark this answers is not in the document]
slavistan: It's more than just the `@type` field. The payloads for the attribute succession routes require only a subset of the fields a of a full attribute (e.g. when succeeding a relationship attribute the `key` property must not be specified, while it's mandatory when creating a relationship attribute).

What I could do in order to explicitly reference the existing attribute types is this

```ts
import { ConnectorRelationshipAttribute } from "../ConnectorAttribute";

export interface SucceedRelationshipAttributeAndNotifyPeerRequest {
    predecessorId: string;
    successorContent: {
        value: ConnectorRelationshipAttribute["value"];
        validFrom?: ConnectorRelationshipAttribute["validFrom"];
        validTo?: ConnectorRelationshipAttribute["validTo"];
    };
}
```

although I'm not sure that there's any real benefit. The payloads for creating attribute vs succeeding them are just different, period. We don't have to lump them together.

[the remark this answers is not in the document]
jkoenig134: You could use `Pick` to pick only the three fields?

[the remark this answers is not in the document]
jkoenig134: I **hate** this syntax. `value: ConnectorRelationshipAttribute["value"];`

packages/sdk/src/endpoints/AttributesEndpoint.ts
stnmtz: Would this be my option a)? :D

slavistan: Yepp, that's right. @jkoenig134 We've got a tie here. Which is it to be, a or b?

jkoenig134: as we have no other convenience routes at the moment I'd "vote" with a, but this isn't a democracy, we should discuss this more deeply :D

src/modules/coreHttpApi/controllers/AttributesController.ts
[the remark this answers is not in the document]
slavistan: It's a major pita to work with `unknown`s, but I do see the merits of it. We can use `unknown`s over `any`s, but I rather move this discussion/refactoring into a separate PR. What do you think?

test/attributes.test.ts
slavistan: Possible. We haven't found the culprit yet.

src/modules/coreHttpApi/controllers/AttributesController.ts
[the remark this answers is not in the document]
jkoenig134: yep, it is uncomfortable, but we should use unknown where we have custom logic on the request. I'm ok with doing it in another PR. The errors should now be resolved.

packages/sdk/src/types/attributes/requests/ShareAttributeRequest.ts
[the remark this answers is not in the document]
slavistan: Looks like a relic from an earlier draft. Highly suspicious.

What we do with it depends on whether we keep or remove the `shareAttribute` request/endpoint altogether.

packages/sdk/src/endpoints/AttributesEndpoint.ts
slavistan: I'll be waiting for a verdict.

jkoenig134: No, lets talk together like Thomas suggested!

jkoenig134:
@slavistan opens a new one

PR #128, opened 2024-01-25
Feature/attribute succession

slavistan:


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 3.7.0
+
+-> SDK 3.0.0
+
+-   fix: POST /api/v2/Attributes will no longer accept RelationshipAttributes, as this was unintentional and led to wrong behavior
+-   add routes for attribute succession
```
jkoenig134, line 8: maybe list the added routes?

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 3.7.0
+
+-> SDK 3.0.0
+
+-   fix: POST /api/v2/Attributes will no longer accept RelationshipAttributes, as this was unintentional and led to wrong behavior
```
jkoenig134, line 7:
```suggestion
-   fix: `POST /api/v2/Attributes` will no longer accept RelationshipAttributes, as this was unintentional and led to wrong behavior
```

packages/sdk/src/endpoints/AttributesEndpoint.ts
```diff
@@ -2,21 +2,35 @@ import {
     ConnectorAttribute,
     ConnectorAttributes,
     ConnectorResponse,
-    CreateAttributeRequest,
+    CreateIdentityAttributeRequest as CreateRepositoryAttributeRequest,
```
jkoenig134, line 5: why is this aliased (vs renamed in the actual file)?

slavistan:
Local tests run fine on mongo db (`npm run test:local:mongo`), but fail when run on ferret db (`npm run test:local:ferret`).

jkoenig134:
@slavistan same with ci tests. Let's first tackle the review and then look at what could be the root cause.

src/modules/coreHttpApi/controllers/AttributesController.ts
```diff
@@ -1,25 +1,58 @@
...
-    ) {
+    public constructor(@Inject private readonly consumptionServices: ConsumptionServices) {
         super();
     }
 
     @POST
     @Accept("application/json")
-    public async createAttribute(request: any): Promise<Return.NewResource<Envelope>> {
-        const selfAddress = (await this.transportServices.account.getIdentityInfo()).value.address;
-        if (request?.content?.owner !== selfAddress) throw new Error("You are not allowed to create an attribute that is not owned by yourself");
```
jkoenig134, line 20: I think to be really backwards compatible this check should be re-added. I mean - if owner if specified AND is set to not the own address we just delete it.

test/attributes.test.ts
```diff
@@ -49,16 +56,106 @@ describe("Attributes", () => {
...
+        const attribute = await client1.attributes.createRepositoryAttribute({
+            content: {
+                value: {
+                    "@type": "GivenName",
+                    value: "AGivenName"
+                }
+            }
+        });
+        expect(attribute.isSuccess).toBe(true);
+    });
```
jkoenig134, line 70: this is copy paste from the first test (`should create a repository attribute` lowercase)

test/iqlquery.test.ts
```diff
@@ -75,6 +75,7 @@ test("Local IQL Query", async () => {
     for (const e of table) {
         const queryRequest: ExecuteIQLQueryRequest = {
             query: {
+                "@type": "IQLQuery",
```
jkoenig134, line 78: why isn't this optional anymore?

test/lib/testUtils.ts
```diff
@@ -53,6 +57,27 @@ export async function syncUntilHasMessages(client: ConnectorClient, expectedNumb
...
 }
 
+export async function syncUntilHasMessageWithRequest(client: ConnectorClient, requestId: string): Promise<ConnectorMessage> {
+    const filterRequestMessagesByRequestId = (syncResult: ConnectorSyncResult) => {
+        return syncResult.messages.filter((m: ConnectorMessage) => (m.content as any)["@type"] === "Request" && (m.content as any).id === requestId);
+    };
+    const syncResult = await syncUntil(client, (syncResult) => filterRequestMessagesByRequestId(syncResult).length !== 0);
+    return filterRequestMessagesByRequestId(syncResult)[0];
+}
+export async function syncUntilHasMessageWithResponse(client: ConnectorClient, requestId: string): Promise<ConnectorMessage> {
```
jkoenig134, line 67: missing newline

test/attributes.test.ts
```diff
@@ -141,33 +233,49 @@ describe("Execute AttributeQueries", () => {
...
         });
+
         expect(executeRelationshipAttributeQueryResult).toBeSuccessful(ValidationSchema.ConnectorAttribute);
 
-        expect(executeRelationshipAttributeQueryResult.result.content.value.value).toBe("AString");
+        expect(executeRelationshipAttributeQueryResult.result.content.value.value).toBe("AGivenName");
+
+        expect(executeRelationshipAttributeQueryResult2).toBeSuccessful(ValidationSchema.ConnectorAttribute);
+
+        expect(executeRelationshipAttributeQueryResult2.result.content.value.value).toBe("AGivenName");
```
jkoenig134, line 279: I don't think we need to duplicate the query here. Querying at one connector completely satisfies this test.

test/attributes.test.ts
```diff
@@ -49,16 +56,106 @@ describe("Attributes", () => {
...
+                value: {
+                    "@type": "GivenName",
+                    value: "AGivenName"
+                }
+            }
+        });
+        expect(attribute.isSuccess).toBe(true);
+    });
+
+    test("Should notify peer about Identity Attribute Succession", async () => {
```
jkoenig134, line 72: I'd like to see that to peer is actually notified here

test/attributes.test.ts
```diff
@@ -49,16 +56,106 @@ describe("Attributes", () => {
...
+            }
+        });
+        expect(successionResponse.isSuccess).toBe(true);
+        const notificationResult = await client1.attributes.notifyPeerAboutIdentityAttributeSuccession(successionResponse.result.successor.id, {
+            peer: client2Address
+        });
+        expect(notificationResult.isSuccess).toBe(true);
+    });
+
+    test("should succeed a Repository Attribute", async () => {
```
jkoenig134, line 101: I'd like to see that the attribute is actually succeeded here

test/attributes.test.ts
```diff
@@ -49,16 +56,106 @@ describe("Attributes", () => {
...
+                    value: "AGivenName"
+                },
+                tags: ["content:edu.de"]
+            }
+        });
+
+        expect(succeedAttributeResponse.isSuccess).toBe(true);
+    });
+
+    test("Should succeed a Relationship Attribute", async () => {
```
jkoenig134, line 127: see above

jkoenig134 (approved):

jkoenig134:
It seems like the tests fail against ferretdb when ferretdb is running against postgres. This is REALLY bad, but shouldn't block us for now.

test/attributes.test.ts
```diff
@@ -49,16 +56,106 @@ describe("Attributes", () => {
...
+        const successionResponse = await client1.attributes.succeedAttribute(ownUnsharedRepositoryAttribute.result.id, {
+            successorContent: {
+                value: {
+                    "@type": "GivenName",
+                    value: "ANewGivenName"
+                }
+            }
+        });
+        expect(successionResponse.isSuccess).toBe(true);
+        const notificationResult = await client1.attributes.notifyPeerAboutIdentityAttributeSuccession(successionResponse.result.successor.id, {
```
Magnus-Kuhn, line 95: consistent variable naming throughout the tests: e. g. succeedAttribute returns a successionResponse, a succeedAttributeResponse and a result, notifyPeerAbout... returns a notificationResult

test/iqlquery.test.ts
sebbi08: @J-Gann do you know that ?

jkoenig134 (changes requested):

jkoenig134 (dismissed):
LGTM, @stnmtz please also leave a review.

stnmtz:
Found some things, I'm on it reviewing

test/lib/customMatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+            received = received.toJSON();
+        }
+        if (expected instanceof SerializableBase) {
+            expected = expected.toJSON();
+        }
+
+        const receivedClone = JSON.parse(JSON.stringify(received));
+        const expectedClone = JSON.parse(JSON.stringify(expected));
+
+        excludes.forEach((exclud) => {
```
stnmtz, line 16: exclude?

test/lib/customMatcher.ts
```diff
@@ -0,0 +1,59 @@
...
+                "Received",
+                this.expand ?? true
+            )}`;
+        }
+
+        return { message: () => message, pass };
+    }
+});
+
+export {};
```
stnmtz, line 51: export {} ? Why is that? I would for a comment.

src/modules/coreHttpApi/openapi.yml
```diff
@@ -431,6 +442,146 @@ paths:
...
+            schema:
+              additionalProperties: false
+              properties:
+                peer:
+                  $ref: "#/components/schemas/Address"
+
+      responses:
+        200:
+          description: |
+            "Success. Returns the successor and predecessor own shared identity
```
stnmtz, line 547: Own shared identity attributes is mentioned here, would it be better to use repository attributes?

src/modules/coreHttpApi/openapi.yml
```diff
@@ -431,6 +442,146 @@ paths:
...
+                      $ref: "#/components/schemas/IdentityAttributeContent_tags"
+                    validFrom:
+                      $ref: "#/components/schemas/AttributeContent_validFrom"
+                    validTo:
+                      $ref: "#/components/schemas/AttributeContent_validTo"
+
+      responses:
+        200:
+          description: |
+            Success. Succeeding a Relationship Attribute will return an
```
stnmtz, line 482: > Succeeding a Relationship Attribute will return an

Use present tense. predecessor/successor is missing, why "additional" ID?

src/modules/coreHttpApi/openapi.yml
```diff
@@ -431,6 +442,146 @@ paths:
         403:
           $ref: "#/components/responses/Forbidden"
 
+  /api/v2/Attributes/{predecessorId}/Succeed:
+    post:
+      operationId: succeedAttribute
+      description: Succeeds either a Relationship Attribute or a Repository Attribute.
```
stnmtz, line 448: Casing of Repository Attribute vs. repository attribute below

src/modules/coreHttpApi/openapi.yml
```diff
@@ -431,6 +442,146 @@ paths:
...
+                      $ref: "#/components/schemas/AttributeValue"
+                    tags:
+                      $ref: "#/components/schemas/IdentityAttributeContent_tags"
+                    validFrom:
+                      $ref: "#/components/schemas/AttributeContent_validFrom"
+                    validTo:
+                      $ref: "#/components/schemas/AttributeContent_validTo"
+
+      responses:
+        200:
```
stnmtz, line 480: Why 200 on POST?

src/modules/coreHttpApi/controllers/AttributesController.ts
```diff
@@ -7,19 +8,59 @@ import { BaseController } from "../common/BaseController";
...
     }
 
     @POST
     @Accept("application/json")
-    public async createAttribute(request: any): Promise<Return.NewResource<Envelope>> {
+    public async createRepositoryAttribute(request: any): Promise<Return.NewResource<Envelope>> {
         const selfAddress = (await this.transportServices.account.getIdentityInfo()).value.address;
-        if (request?.content?.owner !== selfAddress) throw new Error("You are not allowed to create an attribute that is not owned by yourself");
+        if (request?.content?.owner && request?.content?.owner !== selfAddress) {
+            throw new ApplicationError("error.connector.attributes.cannotCreateNotSelfOwnedAttribute", "You are not allowed to create an attribute that is not owned by yourself");
```
stnmtz, line 22: cannotCreateNotSelfOwnedAttribute => cannotCreateNotSelfOwnedRepositoryAttribute

packages/sdk/src/types/attributes/IQLQuery.ts
```diff
@@ -1,5 +1,5 @@
 export interface IQLQuery {
-    "@type"?: "IQLQuery";
+    "@type": "IQLQuery";
```
stnmtz, line 2: As Julian asked in the tests, why no longer optional?

README_dev.md
```diff
@@ -4,7 +4,8 @@
 2. download and install [Node JS](https://nodejs.org/en/download/)
 3. install the npm packages `typescript` and `cpx` **globally**
 4. Optional: install the VS Code extension [Tasks](https://marketplace.visualstudio.com/items?itemName=actboy168.tasks)
-5. run `npm i`
+5. Optional: restarting the Connector using the VSCode task `More Tasks > Restart` requires the tool socat. Make sure to install it on your system. This can be done for example using `apt-get install socat` on Ubuntu and `brew install socat` using Homebrew.
```
stnmtz, line 7: Why optional? Why "make sure to install it ..." when it is optional?

stnmtz (changes requested):

README_dev.md
sebbi08: I think it is optional as it is only required for manual restart. By default, the restart will be handled by nodemon

jkoenig134: I mean, partly true: nodemon watches the files and restarts the server on changes. But sometimes (e.g. when nodemon fails to startup the server) we need to have a manual way to do it.

But yeah, it's very optional b/c it's only needed soooooometimes :D

test/lib/customMatcher.ts
jkoenig134: this is for typescript to actually compile this. You instantly get a compiler error when removed so I vote against this comment :)

src/modules/coreHttpApi/openapi.yml
sebbi08: The ID is only set when a relationship attribute is suceeded as this will directly be send to the peer via notification

test/lib/customMatcher.ts
sebbi08: actually not needed anymore seems to be an update in ts or something else :/

packages/sdk/src/types/attributes/requests/CreateIdentityAttributeRequest.ts
```diff
@@ -0,0 +1,10 @@
+import { ConnectorAttributeValue } from "../ConnectorAttribute";
+
+export interface CreateRepositoryAttributeRequest {
```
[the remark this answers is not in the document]
sebbi08: done

README_dev.md
```diff
@@ -4,7 +4,8 @@
 2. download and install [Node JS](https://nodejs.org/en/download/)
 3. install the npm packages `typescript` and `cpx` **globally**
```
slavistan, line 5: What's the point in installing `typescript` and `cpx` globally? Let's add `typescript` and `cpx` as dev dependencies instead.

jkoenig134: cpx had vulnerabilities for a while (edit: still has) and wasn't included in the repo therefore **(leave this as is!)** 

typescript is already included, but WAY MORE better to use when installed globally. If you don't want to have it globally installed just skippit :)

sebbi08: We could use https://github.com/bcomnes/cpx2 as this is maintained and a drop-in replacement

jkoenig134: Ok, sounds great!

jkoenig134: as this is not part of this PR I'll add this in another PR.

packages/sdk/src/types/attributes/IQLQuery.ts
sebbi08: Hi this is a little bit irritating for me.

https://github.com/nmshd/runtime/blob/39b5138435a9b7e137a36d21278c76fa81e0fe33/packages/content/src/attributes/IQLQuery.ts#L33

Here it also is not optional anymore. But in the `from` method the property is omitted

https://github.com/nmshd/runtime/blob/39b5138435a9b7e137a36d21278c76fa81e0fe33/packages/content/src/attributes/IQLQuery.ts#L59

sebbi08: When I try to call the runtime use case and omit the `"@type": "IQLQuery"` i will get an error that it could not be parsed

src/modules/coreHttpApi/openapi.yml
```diff
@@ -431,6 +442,146 @@ paths:
...
+        content:
+          application/json:
+            schema:
+              additionalProperties: false
+              properties:
+                peer:
+                  $ref: "#/components/schemas/Address"
+
+      responses:
+        200:
```
stnmtz, line 545: Again 200 on POST :P

packages/sdk/src/types/attributes/IQLQuery.ts
stnmtz: > When I try to call the runtime use case and omit the "@type": "IQLQuery" i will get an error that it could not be parsed
Ok, that's strange ...

- As we directly deserialize via IQLQuery.from with the Runtime's Usecase (see https://github.com/nmshd/runtime/blob/main/packages/runtime/src/useCases/consumption/attributes/ExecuteIQLQuery.ts#L28 ) we do not need @type: "IQLQuery" within the JSON for any payload to POST .../ExecuteIQLQuery (as ts-serval is automatically using the default type then)
- I guess this is caused by a TypeScript type error b/c the interface of ExecuteIQLQueryRequest seems to be wrong, as it should be possible to omit the @ type definition in the request (because of the first bullet point): https://github.com/nmshd/runtime/blob/main/packages/runtime/src/useCases/consumption/attributes/ExecuteIQLQuery.ts#L10
- I can remember that @slavistan had some problems when building the schemas within the Runtime because of some ugly bugs which might be a reason/cause of not omitting the @ type within the ExecuteIQLQueryRequest interface in the Runtime.
- BUT: I can imagine that using the Connector however all works out good, as all the requests coming in are any's and thus the TypeScript error is ommitted.
- Thus: I can imagine that using the SDK without the @ type: "IQLQuery" definition is working great, and we could remove this change. But we still have an inconsistency within the Runtime's request types of the Usecases.


Furthermore:
- Compared to RelationshipAttributeQuery, we're doing exact the same thing within the Runtime (the @ type is defined within the RelationshipAttributeQuery): https://github.com/nmshd/runtime/blob/39b5138435a9b7e137a36d21278c76fa81e0fe33/packages/content/src/attributes/RelationshipAttributeQuery.ts#L69
- BUT within the type definition of RelationshipAttributeQueryJSON in the Connector SDK, the @ type is optional: https://github.com/nmshd/connector/blob/main/packages/sdk/src/types/attributes/RelationshipAttributeQuery.ts#L2
- So as we're doing the same thing, using the Runtime Usecase ExecuteRelationshipAttributeQuery without an @ type:"RelationshipAttributeQuery" is also doing fine?
- And we might have the same Runtime request type issues for Execute... (or even more?)

J-Gann: I believe the error occurs during validation of the UseCase: https://github.com/nmshd/runtime/blob/main/packages/runtime/src/useCases/consumption/attributes/ExecuteIQLQuery.ts#L15, as the schema of "ExecuteIQLQueryRequest" requires @type to be defined:
https://github.com/nmshd/runtime/blob/main/packages/runtime/src/useCases/common/Schemas.ts#L15948
This validation seems to be missing in the IdentityAttributeQuery: https://github.com/nmshd/runtime/blob/main/packages/runtime/src/useCases/consumption/attributes/ExecuteIdentityAttributeQuery.ts#L15

jkoenig134: I'll wait for this to be merged and will update the code

sebbi08: is optional again

jkoenig134: I've gone a quite different route - recreating what we have in the actual types. Now we can omit it when KNOWING what kind of query we have.

jkoenig134 (dismissed):
LGTM, but please wait for @stnmtz to also approve.

jkoenig134 (dismissed):

jkoenig134:
@stnmtz we will merge this, if you find sth. we can do this afterwards.

PR #134, opened 2024-02-13
refactor/apply consistent attribute nomenclature

Magnus-Kuhn:
use repositoryAttributes and ownSharedIdentityAttributes, update runtime dependency


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@nmshd/connector",
-    "version": "3.7.1",
+    "version": "4.0.0",
```
jkoenig134, line 3: can be a patch bump, you don't change the actual routes.

jkoenig134: don't forget the changelog pls

jkoenig134:
don't forget the version bump for the SDK. (major)

src/modules/coreHttpApi/openapi.yml
jkoenig134, line 1: I think here are MANY changes missing

Magnus-Kuhn: the only other candidate for changing I see is asking for a "peer shared relationship attribute" in the third party relation attribute query, but the context is clear enough imo - calling things generically "attribute" still works

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.7.2
+
+-   upgrade the runtime version to 3.7.2
+-   use the names 'repository attribute' and 'own shared identity attribute'
```
jkoenig134, line 6: this is just internal and should not occur in the changelog

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## 3.7.2
```
jkoenig134, line 3: missing related sdk version (see below on how to add this)

package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
     "name": "@nmshd/connector",
-    "version": "3.7.1",
+    "version": "4.0.0",
```
jkoenig134, line 3: seems like you forgot `npm i` after the version downgrade

src/modules/coreHttpApi/openapi.yml
jkoenig134: yeah, I see, most things were renamed before.

jkoenig134 (changes requested):

CHANGELOG.md
```diff
@@ -2,8 +2,8 @@
 
 ## 3.7.2
 
--   upgrade the runtime version to 3.7.2
--   use the names 'repository attribute' and 'own shared identity attribute'
+-   upgrade the runtime version to 4.0.0
+-   upgrade the sdk version to 4.0.0
```
jkoenig134, line 6: nope, look under `## 3.7.0`, there you can see how to reference the sdk

jkoenig134 (approved):

test/attributes.test.ts
```diff
@@ -120,7 +120,7 @@ describe("Attributes", () => {
             }
         });
         expect(successionResponse.isSuccess).toBe(true);
-        const notificationResponse = await client1.attributes.notifyPeerAboutIdentityAttributeSuccession(successionResponse.result.successor.id, {
+        const notificationResponse = await client1.attributes.notifyPeerAboutRepositoryAttributeSuccession(successionResponse.result.successor.id, {
```
slavistan, line 123: Kindly, leave the name as it was: The notification sent to the peer contains an update to the _own shared identity attribute_, not a repository attribute. (I know .. there's room for three hours of bikeshedding about this particular name. Been there, done that and afterwards we named the runtime use case `notifyPeerAboutIdentityAttributeSuccession`.)

slavistan (dismissed):

test/attributes.test.ts
Magnus-Kuhn: since this concerns a public sdk method, this change would also have to apply to some types and the openAPI specification. We discussed those with a different conclusion yesterday, are you fine with those changing as well?

slavistan: I'm not sure I understand. I meant for you to leave the method's name as `notifyPeerAboutIdentityAttributeSuccession`. Just revert your changes for this particular method name and the names of any related files.

Please elaborate if I missed your point: who discussed what with what conclusion and what changes are you talking about?

jkoenig134: `notifyPeerAboutIdentityAttributeSuccession` is not correct imo. In the runtime the UseCase and facade methods are called `notifyPeerAboutRepositoryAttributeSuccession` so this should map this method.

slavistan: >In the runtime the UseCase and facade methods are called `notifyPeerAboutRepositoryAttributeSuccession`

That change went unnoticed to me. In that case we obviously apply the same naming here. Thus, forget what I said.

PR #137, opened 2024-02-16
Fix/test stability

sebbi08:
This pull request fixes the stability of the tests by waiting for events of message handling. It also updates the version to 3.7.4.


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@nmshd/connector",
-    "version": "3.7.3",
+    "version": "3.7.4",
```
jkoenig134, line 3: don't forget to run `npm i` when you undo this.

test/lib/Launcher.ts
```diff
@@ -52,15 +52,15 @@ export class Launcher {
...
         }
 
         await Promise.all(ports.map(waitForConnector));
         return clients;
     }
 
-    private randomString(): string {
+    public static randomString(): string {
         return Math.random().toString(36).substring(7);
     }
```
jkoenig134, line 65: you should use the method from the other branch

jkoenig134 (changes requested):
some comments, but we should also tackle the webhookServer.

test/lib/Launcher.ts
jkoenig134: and don't make this static

PR #140, opened 2024-03-06
chore/improve readme and ignore local config file

sebbi08:


README_dev.md
```diff
@@ -106,7 +106,7 @@ npm run test:local -- testSuiteName
           "platformClientId": "...",
           "platformClientSecret": "..."
       },
-      "database": { "driver": "lokijs", "folder": "./" },
+      "database": { "driver": "lokijs", "folder": "./", "dbName": "local" },
```
jkoenig134, line 109: why? we have a default for that.

sebbi08: nop for me, it was not working without that config the default is only in the docker compose file I suppose

jkoenig134: image
we have a default, we have to check elsewhere why it doesn't pulls this default for you.

jkoenig134: (but for now, please remove it from here)

jkoenig134 (approved):

PR #141, opened 2024-03-06
test/checks for query and url parameter definition

sebbi08:


package.json
```diff
@@ -60,6 +60,7 @@
         }
     },
     "dependencies": {
+        "@apidevtools/swagger-parser": "^10.1.0",
```
jkoenig134, line 63: why is this a dependency? this shouldn't be used in productive code.

jkoenig134: please don't forget to do an `npm i` after moving it.

jkoenig134 (approved):

jkoenig134 (approved):

PR #146, opened 2024-03-18
Feature/add routes to delete attributes

sebbi08:


sebbi08:
Requires https://github.com/nmshd/runtime/pull/42 to be merged

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## 3.10.0
+
```
jkoenig134, line 4: missing runtime version in the changelog

.dev/compose.yml
```diff
@@ -35,7 +35,7 @@ services:
     environment:
       - CUSTOM_CONFIG_LOCATION=/config.json
       - SYNC_ENABLED=${SYNC_ENABLED:-false}
-      - DATABASE_NAME=${DATABASE_NAME_PREFIX}-2
+      - DATABASE_NAME=${DATABASE_NAME_PREFIX:-connector}-2
```
jkoenig134, line 38: looks like this was committed by accident

package.json
```diff
@@ -15,7 +15,7 @@
         "lint": "npm run lint:tsc && npm run lint:prettier && npm run lint:eslint",
         "lint:prettier": "prettier --check ./src ./test ./packages/*/src",
         "lint:eslint": "eslint --ext .ts ./src ./test ./packages/*/src",
-        "lint:tsc": "npm run build --workspaces && tsc -p test/tsconfig.json --noEmit",
+        "lint:tsc": "npm run build --workspace packages/sdk && tsc -p test/tsconfig.json --noEmit",
```
jkoenig134, line 18: as there is the plan to have more packages in the future you should undo that

packages/sdk/package.json
```diff
@@ -33,7 +33,8 @@
     "dependencies": {
         "axios": "^1.6.7",
         "form-data": "^4.0.0",
-        "qs": "^6.11.2"
+        "qs": "^6.11.2",
+        "@nmshd/runtime": "^4.3.7"
```
jkoenig134, line 37: we don't use the runtime in the SDK. it's too bloated

packages/sdk/src/endpoints/IncomingRequestsEndpoint.ts
```diff
@@ -16,7 +17,10 @@ export class IncomingRequestsEndpoint extends Endpoint {
         return await this.put(`/api/v2/Requests/Incoming/${requestId}/CanAccept`, request);
     }
 
-    public async accept(requestId: string, request: AcceptIncomingRequestRequest): Promise<ConnectorResponse<ConnectorRequest>> {
+    public async accept(requestId: string, request: Omit<AcceptIncomingRequestRequestNew, "requestId">): Promise<ConnectorResponse<ConnectorRequest>> {
+        return await this.put(`/api/v2/Requests/Incoming/${requestId}/Accept`, request);
+    }
+    public async acceptold(requestId: string, request: AcceptIncomingRequestRequestOld): Promise<ConnectorResponse<ConnectorRequest>> {
         return await this.put(`/api/v2/Requests/Incoming/${requestId}/Accept`, request);
     }
```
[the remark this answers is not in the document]
jkoenig134: runtime is too bloated. just remove the acceptold method and use the new one if it's better IMO

packages/sdk/src/types/notification/ConnectorNotificationId.ts
```diff
@@ -0,0 +1,3 @@
+export interface ConnectorNotificationId {
+    notificationId: string;
+}
```
jkoenig134, line 3: the naming of this interface is not good.

It's used to describe a Response that is including  a notification id. Instead you should create one response object for every Method that returns such a response.

src/modules/coreHttpApi/controllers/AttributesController.ts
```diff
@@ -214,6 +214,34 @@ export class AttributesController extends BaseController {
...
+
+    @DELETE
+    @Path("/Peer/Shared/:id")
+    public async deletePeerSharedAttributeAndNotifyOwner(@PathParam("id") attributeId: string): Promise<Envelope> {
+        const result = await this.consumptionServices.attributes.deletePeerSharedAttributeAndNotifyOwner({ attributeId });
+        return this.ok(result);
+    }
+
+    @DELETE
+    @Path("/Third/Party/:id")
```
jkoenig134, line 232: I am a bit confused, why not `/ThirdParty/:id`? The words third and party do not make any sense when not combined in this context

src/modules/coreHttpApi/controllers/AttributesController.ts
```diff
@@ -214,6 +214,34 @@ export class AttributesController extends BaseController {
...
+    public async deleteThirdPartyAttributeAndNotifyOwner(@PathParam("id") attributeId: string): Promise<Envelope> {
```
jkoenig134, line 233:
```suggestion
    public async deleteThirdPartyAttributeAndNotifyPeer(@PathParam("id") attributeId: string): Promise<Envelope> {
```

test/attributes.test.ts
```diff
@@ -22,7 +30,7 @@ let client1Address: string;
 let client2Address: string;
 
 beforeAll(async () => {
-    [client1, client2] = await launcher.launch(2);
+    [client1, client2] = await launcher.launch(3);
```
jkoenig134, line 33: you launch 3 but only use 3, any reason for that?

test/iqlquery.test.ts
jkoenig134, line 1: this change is related to another issue, right?

test/lib/Launcher.ts
```diff
@@ -37,7 +37,7 @@ export class Launcher {
 
         for (let i = 0; i < count; i++) {
             const port = await getPort();
-            const accountName = await this.randomString();
+            const accountName = `${i + 1}-${await this.randomString()}`;
```
jkoenig134, line 40: ?

packages/sdk/src/endpoints/IncomingRequestsEndpoint.ts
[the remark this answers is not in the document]
jkoenig134: with use the new one i mean "update the old one that it gets like the other type"

jkoenig134 (changes requested):

.dev/compose.yml
sebbi08: Na I just copied it from connector1 without this, connector 2 won't start if no DATABASE_NAME_PREFIX is provided

jkoenig134: IC!

test/lib/Launcher.ts
sebbi08: Oh, somehow my code comment got lost it is there to improve debugging the tests when accessing the DB to see what client is related to which DB

packages/sdk/src/endpoints/IncomingRequestsEndpoint.ts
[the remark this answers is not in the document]
sebbi08: Yea, but the types in the SDK are just wrong. And are not checked against the real accepted types of the runtime. Also, if there is a new subclass in the runtime that comes from time to time this is not noticeable in the SDK.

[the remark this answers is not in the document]
jkoenig134: cool, when they are wrong you don't have to provide an "old" and "new" and can just update the wrong types.

test/lib/Launcher.ts
jkoenig134: Oh, I see. When debugging one test suite at once this gets really handy, right? I don't think we need a comment for that, I'm ok with the change.

packages/sdk/src/endpoints/IncomingRequestsEndpoint.ts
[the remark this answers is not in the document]
sebbi08: Yea but the new type is the type of the @nmshd/runtime

test/iqlquery.test.ts
sebbi08: ah, yea accidentally committed it will move it to another PR

packages/sdk/src/endpoints/IncomingRequestsEndpoint.ts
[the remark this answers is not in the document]
jkoenig134: what's the problem with updating the old type? I don't get that...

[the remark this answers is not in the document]
sebbi08: There is no problem but for, i don't know how long due to the mono repo. There was no way in the SDK to create a
AcceptReadAttributeRequestItemParametersWithExistingAttributeJSON
AcceptProposeAttributeRequestItemParametersWithExistingAttributeJSON
AcceptFreeTextRequestItemParametersJSON

And I think if we really stay with copying (recreating) all the types from runtime in the SDK, this will appear in many other places.

How about we bundle that SDK and import the runtime types so that not all the runtime code is used?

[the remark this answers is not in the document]
jkoenig134: It's the plan to have a package only containing types that is used by the runtime and the sdk - but we aren't there yet. for now

1. allow any / or make the type way more open
2. copy the types for now

jkoenig134:
you are missing an sdk version bump. Don't forget to also add this to the changelog.

test/attributes.test.ts
sebbi08: nop just a missing undo from an earlier implementation

src/modules/coreHttpApi/controllers/AttributesController.ts
```diff
@@ -214,6 +214,34 @@ export class AttributesController extends BaseController {
...
+    @DELETE
+    @Path("/Peer/Shared/:id")
+    public async deletePeerSharedAttributeAndNotifyOwner(@PathParam("id") attributeId: string): Promise<Envelope> {
+        const result = await this.consumptionServices.attributes.deletePeerSharedAttributeAndNotifyOwner({ attributeId });
+        return this.ok(result);
+    }
+
+    @DELETE
+    @Path("/ThirdParty/:id")
+    public async deleteThirdPartyAttributeAndNotifyPeer(@PathParam("id") attributeId: string): Promise<Envelope> {
```
jkoenig134, line 233: again naming: deleteThirdPartyOwnedRelationshipAttributeAndNotifyPeer
```suggestion
    public async deleteThirdPartyOwnedRelationshipAttributeAndNotifyPeer(@PathParam("id") attributeId: string): Promise<Envelope> {
```

test/attributes.test.ts
```diff
@@ -1,6 +1,7 @@
 import { DataEvent } from "@js-soft/ts-utils";
 import { ConnectorAttribute, ConnectorRelationshipAttribute, ConnectorResponse } from "@nmshd/connector-sdk";
-import { SuccessionEventData } from "@nmshd/runtime";
+import { DeletionStatus, IncomingRequestStatusChangedEvent, ThirdPartyOwnedRelationshipAttributeDeletedByPeerEvent } from "@nmshd/consumption";
```
jkoenig134, line 3: why do you have to import events from consumption?

sebbi08: To get type completion in the test event waiter. By default, it will use the event from ts-utils and this has no other property than the namespace.

jkoenig134: yeah, but events from consumptions should not be used here, instead use the runtime events if **really** necessary.

jkoenig134 (approved):

PR #166, opened 2024-05-29
Implement the new Relationship API

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.

# Description

-   New route to revoke relationship
    -   `PUT /api/v2/Relationships/{id}/Revoke`
-   Remove relationship changes in favor of new routes

    -   REMOVED `PUT /api/v2/Relationships/{id}/Changes/{changeId}/Reject`
    -   NEW `PUT /api/v2/Relationships/{id}/Reject`
    -   REMOVED `PUT /api/v2/Relationships/{id}/Changes/{changeId}/Accept`
    -   NEW `PUT /api/v2/Relationships/{id}/Accept`

-   update runtime version to `5.0.0-alpha.1`


jkoenig134:
We need to publish pre-release npm packages for the libs to merge this, right?

test/messages.test.ts
```diff
@@ -93,7 +93,7 @@ describe("Messaging", () => {
 });
 
 describe("Message errors", () => {
-    const fakeAddress = "id1PNvUP4jHD74qo6usnWNoaFGFf33MXZi6c";
+    const fakeAddress = "did:e:fakeAddress:dids:0123456789012345678912";
```
jkoenig134, line 96:
```suggestion
    const fakeAddress = "did:e:a-domain:dids:0000000000000000000000";
```
we should use this everywhere we use a valid but fake address.

src/modules/coreHttpApi/openapi.yml
jkoenig134, line 1: I'm missing revoke here

sebbi08: This is only "old" functionality updated to V5 the Termination will be handled in another PR

src/modules/autoAcceptRelationshipCreationChanges/AutoAcceptRelationshipCreationChangesModule.ts
```diff
@@ -39,10 +37,11 @@ export default class AutoAcceptRelationshipCreationChangesModule extends Connect
 
     private isIncomingPendingRelationshipCreationChange(event: RelationshipChangedEvent) {
         const data = event.data;
-        if (data.changes.length !== 1) return false;
+        if (data.status !== RelationshipStatus.Pending) return false;
 
-        const creationChange = event.data.changes[0];
-        return creationChange.request.createdBy !== this.currentIdentity && creationChange.status === RelationshipChangeStatus.Pending;
+        const creationChange = event.data.auditLog.find((log) => !log.oldStatus && log.newStatus === RelationshipStatus.Pending);
```
jkoenig134, line 42: I think we can massively simplify this.

1. there is only one entry here when the relationship is pending, so you can still look at index 0 (maybe checking for the length and return if its longer than expected)
2. please don't call it creationChange, instead just auditLogEntry
3. you can then just check for the createdBy for the entry on index 0 like you already do

packages/sdk/src/types/relationships/ConnectorRelationship.ts
```diff
@@ -1,13 +1,32 @@
...
+export interface ConnectorRelationshipAuditLogEntry {
+    createdAt: string;
+    createdBy: string;
+    createdByDevice: string;
+    reason: ConnectorRelationshipAuditLogEntryReason;
+    oldStatus?: ConnectorRelationshipStatus;
+    newStatus: ConnectorRelationshipStatus;
+}
+
+export interface ConnectorRelationshipAuditLog extends Array<ConnectorRelationshipAuditLogEntry> {}
```
jkoenig134, line 22: should be a separate file like it was with the RelationshipChangesBefore

src/modules/coreHttpApi/openapi.yml
jkoenig134: revoke is not connected to the termination in any way.

jkoenig134: (the revoke I am talking about here)

sebbi08: I cannot find any reference to revoke in the connector.

jkoenig134: I dunno why we had no revoke on relationship changes, but IMO we should provide it now.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@nmshd/connector",
-    "version": "3.11.2",
```
jkoenig134, line 3: 1. branch says release/v5, please change the branch name
2. as long as the runtime runs under prereleases we have to also publish prereleases here (`4.0.0-alpha.1` see https://github.com/nmshd/runtime/pull/146)

jkoenig134: (same for the SDK)

packages/sdk/src/types/ConnectorResponse.ts
```diff
@@ -38,7 +38,9 @@ export class ConnectorResponse<T> {
 
     public get result(): T {
         if (!this.isSuccess) {
-            throw new Error(`Can't get the value of an error Response. Use 'error' instead. Root error: ${this.error.code} - ${this.error.message}`);
+            throw new Error(
+                `Can't get the value of an error Response. Use 'error' instead. Root error: ${this.error.code} - ${this.error.stacktrace?.length ? this.error.stacktrace.join("\n") : this.error.message}`
```
jkoenig134, line 42: you hide the message if there is a stack trace

sebbi08: Yea you might think so but JS Error stacktraces always contain the message also a bit weird

jkoenig134: IC

jkoenig134 (approved):

PR #202, opened 2024-07-03
[SDK] network error leads to unwanted behavior

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/packages/sdk/src/endpoints/Endpoint.ts
+++ b/packages/sdk/src/endpoints/Endpoint.ts
@@ -48,8 +48,9 @@ export abstract class Endpoint {
         if (httpResponse.status !== expectedStatus) {
             const errorPayload = httpResponse.data.error;
             if (!errorPayload) {
-                const fullUrl = `${httpResponse.request.protocol}//${httpResponse.request.getHeader("host")}${httpResponse.request.path}`;
-                throw new Error(`The http request to connector route '${fullUrl}' failed with status '${httpResponse.status}': ${httpResponse.statusText} ${httpResponse.data}`);
+                throw new Error(
+                    `The http request to connector route '${httpResponse.request.path}' failed with status '${httpResponse.status}': ${httpResponse.statusText} ${httpResponse.data}`
+                );
             }
             return ConnectorResponse.error({
                 id: errorPayload.id,
@@ -71,6 +72,12 @@ export abstract class Endpoint {
         if (httpResponse.status !== 200) {
             // Manually parse data because responseType is "arrayBuffer"
             const errorPayload = JSON.parse(httpResponse.data).error;
+
+            if (!errorPayload) {
+                throw new Error(
+                    `The http request to connector route '${httpResponse.request.path}' failed with status '${httpResponse.status}': ${httpResponse.statusText} ${httpResponse.data}`
+                );
+            }
 
             return ConnectorResponse.error({
                 id: errorPayload.id,
@@ -106,6 +113,12 @@ export abstract class Endpoint {
         const expectedStatus = method === "GET" ? 200 : 201;
         if (httpResponse.status !== expectedStatus) {
             const errorPayload = httpResponse.data.error;
+
+            if (!errorPayload) {
+                throw new Error(
+                    `The http request to connector route '${httpResponse.request.path}' failed with status '${httpResponse.status}': ${httpResponse.statusText} ${httpResponse.data}`
+                );
+            }
 
             return ConnectorResponse.error({
                 id: errorPayload.id,


sebbi08:
When there is a network error between SDK and connector the `errorPayload` is undefined and the SDK crashes with an `can not read property id of undefined `

packages/sdk/src/endpoints/Endpoint.ts
```diff
@@ -47,7 +47,10 @@ export abstract class Endpoint {
 
         if (httpResponse.status !== expectedStatus) {
             const errorPayload = httpResponse.data.error;
```
jkoenig134, line 49: when does this happen? when you call a route that does not exist on the connector?

packages/sdk/src/endpoints/Endpoint.ts
```diff
@@ -47,7 +47,10 @@ export abstract class Endpoint {
...
-
+            if (!errorPayload) {
+                const fullUrl = `${httpResponse.request.protocol}//${httpResponse.request.getHeader("host")}${httpResponse.request.path}`;
```
jkoenig134, line 51: I'd leave out protocol, host and only inject the path

packages/sdk/src/endpoints/Endpoint.ts
```diff
@@ -47,7 +47,10 @@ export abstract class Endpoint {
...
+                throw new Error(`Http request to connector route ${fullUrl} failed with: ${httpResponse.status} ${httpResponse.statusText} ${httpResponse.data}`);
```
jkoenig134, line 52:
```suggestion
                throw new Error(`Http request to connector route '${fullUrl}' failed with status '${httpResponse.status}': ${httpResponse.statusText} ${httpResponse.data}`);
```

packages/sdk/src/endpoints/Endpoint.ts
jkoenig134: oh, I didn't though about that.

packages/sdk/src/endpoints/Endpoint.ts
sebbi08: As this is mostly logged when connecting to the wrong server/port, I would leave it in there.

jkoenig134: as this is user specified information, please leave it out.

jkoenig134: The more I think about it I highly doubt we even need the path. The connector always responds with an error object (even for 404s).

slavistan: If you keep the host use `.request.host`, instead of retrieving it from the request header.

jkoenig134 (approved):

packages/sdk/src/endpoints/Endpoint.ts
slavistan: >when you use a non connect web server

You wouldn't get an HTTP response but an exception instead in that case and `makeResult` would never execute (for example, coming from a call to `Endpoint.post()`). If you want to catch socket hangups etc. you'd have to wrap the HTTP requests in a `try/catch` block and catch `AxiosError`s. `makeResult` is only ever entered after a completed HTTP request/response cycle.

jkoenig134: Actually it depends on how you configure axios.

slavistan: Unless I'm mistaken, axios uses exceptions to signal errors on the network level (socket hangups, unknown hosts, timeouts, ...) and this behavior cannot be changed. So if we want the SDK to gracefully handle these errors, we have to wrap the calls to `this.httpClient.{get,post, ...}()` inside _Endpoint.ts_.

What can be configured though is to not raise an exception for non-2xx and non-3xx HTTP Response codes, which it does by default (and the SDK does override this default behavior, see [here](https://github.com/nmshd/connector/blob/f3a219d13eaae672f3d188db81809f7413bcbd9d/packages/sdk/src/ConnectorClient.ts#L38)). May you're referring to this?

jkoenig134: Yeah, can be that I confused that. But then I am a bit confused about this PR. I don't see the usecase if network errors are already thrown errors.

sebbi08: Hm, yea I rechecked @slavistan is right, real network errors are thrown. This code would only be run when running against a non connector web server or also a proxy that sends some error data back that is not in the connector format.

jkoenig134: But I think this is still ok.

PR #206, opened 2024-07-05
Test would not stop when there is an error in the connector startup

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.

Due to the introduction of the webhook test server, jest would not stop the process because of open handlers when there is an error in the connector startup.


jkoenig134:
Just for me to clarify - this is only fixing the issue of not detecting fast enough, that the connector exited, right? B/c after waiting for the specified time there will be no port open and the test will fail.

sebbi08:
> Just for me to clarify - this is only fixing the issue of not detecting fast enough, that the connector exited, right? B/c after waiting for the specified time there will be no port open and the test will fail.

Yea, the tests will fail, and the connector is closed, but the webhook test server is still open. And this will treigger the open hook detection of jest and then jest will just not stop.

jkoenig134:
Ehrm, then this is more a problem with webhook servers not getting cleaned up where this should happen. I'll look into it.

test/lib/Launcher.ts
```diff
@@ -105,10 +127,20 @@ export class Launcher {
 
     public stop(): void {
         this._processes.forEach((p) => {
-            p.connector.on("exit", () => {
-                p.webhookServer?.close();
-            });
-            p.connector.kill();
```
jkoenig134, line 111: this code made no sense. All you want to do here is

1. close the webhookServer
2. kill the connector

why not just 
```ts
p.webhookServer?.close();
p.connector.kill();
```

IMO your stopClient method is also overkill.

sebbi08: The problem is that between killing the connector and closing the webhook server, there could be some event send. And if the webhook server is closed before, this would create an error in the connector

jkoenig134: But that's not problem, we don't care about any issues here. Additionally you can just call the methods the other way round.

jkoenig134:
@sebbi08 why did you close this? Please never close without a message.

sebbi08:
> @sebbi08 why did you close this? Please never close without a message.

i did not remember to do so :/

jkoenig134 (approved):

PR #207, opened 2024-07-05
Add script to copy local runtime

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.


.dev/scripts/copyRuntime.sh
```diff
@@ -0,0 +1,30 @@
...
+
+source .env
+CONNECTORPATH=$(pwd)
+
+if [ -z "$RUNTIME_PATH" ]; then
+    echo "Please provide the path via the .env in the root folder file"
+    exit 1
+fi
+
+currentPackageName=$(jq .name package.json 2>/dev/null | tr -d '"') || ""
```
slavistan, line 11: You can get unquoted ("raw") output with `jq -r`, i.e.

```sh
jq -r .name package.json 
```

.dev/scripts/copyRuntime.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+source .env
+CONNECTORPATH=$(pwd)
```
slavistan, line 4: The connector path would be 

```sh
# full path stripped of three path segments
CONNECTORPATH="$(dirname "$(dirname "$(dirname "$(realpath "$0")")")")"
```

`$(pwd)` only works from inside the connector's root directory.
```

File: .dev/scripts/copyRuntime.sh [hunk 3]
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+source .env
```
slavistan, line 3: I (and some others) use [autoenv](https://github.com/Tarrasch/zsh-autoenv) to setup my projects' environment so I don't want to be forced into using another file to load environment variables.

I'd just check whether the environment variables your script needs are set without providing a source. If you insist on keeping the _.env_ file, only source it if it exists, pretty please.

```sh
# needs to be in connector root directory (needs a `cd CONNECTORPATH`)
[ -f .env ] && source .env
```

.dev/scripts/copyRuntime.sh
```diff
@@ -0,0 +1,30 @@
...
+if [ -z "$currentPackageName" ] || [ "$currentPackageName" != "@nmshd/connector" ]; then
+    echo "Please run this script from the root of the connector"
+    exit 1
+fi
+
+## declare an array variable
+declare -a arr=("consumption" "content" "iql" "runtime" "transport")
+
+## now loop through the above array
+for package in "${arr[@]}"; do
```
slavistan, line 21: Unnecessary array. You can just list the elements

```bash
for package in consumption content iql runtime transport; do
    # ...
done
```

.dev/scripts/copyRuntime.sh
```diff
@@ -0,0 +1,30 @@
...
+    cd "$RUNTIME_PATH/packages/$package" || exit
+    packageName=$(jq .name package.json | tr -d '"')
+    # printf "Copying $packageName to $currentPackageName \n"
+    cp dist $CONNECTORPATH/node_modules/$packageName -r
```
slavistan, line 25: Dumb question: can't you just symlink the directories? Would make the nodemon script obsolete, too.

sebbi08: Sadly no, in theories it is working, but the linked packages would use their own node_modules folder. And then we run into problems with the typescript-IOC (dependency injection) package because there are two instances of it, the dependency injection from the runtime to the connector is not working anymore.

slavistan: Hm .. weird that it would work with a copy then. Would you mind showing me this on Monday? Maybe we'll figure something out.

slavistan (dismissed):

.dev/scripts/linkRuntime.sh
```diff
@@ -0,0 +1,34 @@
...
+        mkdir -p "$CONNECTORPATH/node_modules/$pkgName"
+        cp -r "$runtimePkgPath/dist" "$runtimePkgPath/src" "$CONNECTORPATH/node_modules/$pkgName"
+    done
+    echo "done!"
+}
+
+case "$1" in
+    "") find $RUNTIME_PATH/packages/*/dist | entr -dd -r -s "$0 _copyRuntime" ;;
+    *) "$@" ;;
+esac
```
jkoenig134, line 34: missing newline

jkoenig134 (changes requested):
rest LGTM.

package.json
```diff
@@ -23,6 +23,8 @@
         "sdk:schemas:watch": "npm run --workspace packages/sdk build:schemas:watch",
         "copy-additional-files": "cpx \"src/**/*.{json,yml,ico}\" dist",
         "copy-additional-files:watch": "cpx \"src/**/*.{json,yml,ico}\" dist --watch",
+        "link-runtime": "bash .dev/scripts/linkRuntime.sh",
+        "unlink-runtime": "rm -rf node_modules/@nmshd && npm install",
```
jkoenig134, line 27: Please run

```bash
npx sort-package-json
```

Also, shouldn't these packages be called `(un)link-local-runtime`?

sebbi08: `npx sort-package-json` had no effect, it only sorts the dependencies.

For the naming I would like to keep it as the only reason for linking is the local usage

sebbi08: `npx sort-package-json` will not work as we have `npm-run-all` installed and that relies on script order

jkoenig134: sort-package-json typically also sorts scripts

jkoenig134: but yeah, for me that also doesn't work. I don't get you npm-run-all command, for me that is not the case but we can close this now.

jkoenig134 (approved):

PR #222, opened 2024-07-31
Add correlation to all requests and the logger

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.


sebbi08:
@jkoenig134 should we write a test for this ?

jkoenig134:
It is possible to test it?

sebbi08:
> It is possible to test it?

yes :D

sebbi08:
@jkoenig134 this should be ready for review

jkoenig134:
I Marke this as draft b/c I am not happy with the tests AND this needs further work from the runtime before it's really useful.

package.json
```diff
@@ -134,7 +134,7 @@
         "@types/yamljs": "^0.2.34",
         "c8": "^10.1.2",
         "cpx2": "^7.0.1",
-        "eslint": "^8.57.0",
+        "eslint": "^9.9.1",
```
jkoenig134, line 137: no .....

src/modules/sse/SseModule.ts
```diff
@@ -88,11 +89,12 @@ export default class SseModule extends ConnectorRuntimeModule<SseModuleConfigura
...
-        if (syncResult.isError) {
-            this.logger.error(syncResult);
-            return;
-        }
+        await correlationIdLibrary.withId(async () => {
+            const syncResult = await services.transportServices.account.syncEverything();
+            if (syncResult.isError) {
+                this.logger.error(syncResult);
+            }
+        });
```
jkoenig134, line 97: maybe this should happen in all the UseCases of the runtime to not have to think about to do it here?

jkoenig134: I think this would affect https://github.com/nmshd/runtime/pull/268 so please wait to merge this until this discussion is resolved.

sebbi08: The problem with doing it in the runtime is that the error log from the connector would not have that correlation ID as it is in the parent scope.

sebbi08: But in general we could add it in the generic execute handler of the runtime and check if there is already one set and reuse this ID

jkoenig134: Hmm, that's bad. But we could do it ALSO in the runtime, right? Then if it's defined it would pick up the ID defined here, and if not create one.

package.json
```diff
@@ -98,6 +99,7 @@
         "mqtt": "^5.10.1",
         "multer": "^1.4.5-lts.1",
         "nconf": "0.12.1",
+        "nyc": "^17.0.0",
```
jkoenig134, line 102: is this required for correlation ids?

sebbi08: Nop don't know how this got there :/ will remove it

package.json
```diff
@@ -113,14 +114,14 @@
         "@js-soft/eslint-config-ts": "1.6.9",
         "@js-soft/license-check": "1.0.9",
         "@nmshd/connector-sdk": "*",
-        "@nmshd/content": "5.4.1",
-        "@nmshd/core-types": "5.4.1",
+        "@nmshd/content": "5.5.0",
```
jkoenig134, line 117: this has to be bumped in the sdk too

please run `ncu -i -ws` in monorepos to not miss the sdk.

src/modules/webhooks/WebhooksModule.ts
```diff
@@ -44,8 +45,11 @@ export default class WebhooksModule extends ConnectorRuntimeModule<WebhooksModul
 
         try {
             this.logger.debug(`Sending request to webhook '${url}' for trigger '${trigger}'.`);
-
-            const response = await this.axios.post(url, payload, { headers: webhook.target.headers });
+            const correlationId = correlator.getId();
```
jkoenig134, line 48: where does this correlationId comes from? as far as I see it it will always be undefined.

sebbi08: Nop the webhook module will be triggered from the runtime events and therefor is most of the time in a call stack of an usecase.

jkoenig134: have you tested this?

jkoenig134: it should be easy to also test this using tests as we already use webhooks for eventing

src/modules/webhooks/WebhooksModule.ts
```diff
@@ -45,7 +46,9 @@ export default class WebhooksModule extends ConnectorRuntimeModule<WebhooksModul
         try {
             this.logger.debug(`Sending request to webhook '${url}' for trigger '${trigger}'.`);
 
-            const response = await this.axios.post(url, payload, { headers: webhook.target.headers });
+            const correlationId = correlator.getId();
+            // eslint-disable-next-line @typescript-eslint/naming-convention
+            const response = await this.axios.post(url, payload, { headers: { ...webhook.target.headers, "x-correlation-id": correlationId } });
```
jkoenig134, line 51: I think it would be safer like this. Also x-headers should be uppercase.

```suggestion
            let headers = webhook.target.headers;
            
            const correlationId = correlator.getId();
            // eslint-disable-next-line @typescript-eslint/naming-convention
            if (correlationId) headers = { ...headers, "X-CORRELATION-ID": correlationId } 
            
            const response = await this.axios.post(url, payload, { headers });
```

sebbi08: Hi I took your fix but timo an I decided to keep the header lower case as headers are anyway case-insensitive and nodejs will anyway convert them to lower case.

test/lib/Launcher.ts
```diff
@@ -130,6 +132,7 @@ export class Launcher {
                 res.status(200).send("OK");
 
                 eventBus.publish(new DataEvent(req.body.trigger, req.body.data));
+                eventBus.publish(new DataEvent("__headers", req.headers));
```
jkoenig134, line 135: why not enhance the data event with the headers?

test/correlationId.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        baseURL: connectorClient1._baseUrl,
+        validateStatus: (_) => true
+    });
+    await establishRelationship(connectorClient1, connectorClient2);
+    account2Address = (await connectorClient2.account.getIdentityInfo()).result.address;
+}, getTimeout(30000));
+
+afterAll(() => launcher.stop());
+describe("should log the correlation id", () => {
+    test("should log the correlation id", async () => {
```
jkoenig134, line 27: I don't see that this test tests any logging?

test/correlationId.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    axiosClient = axios.create({
+        baseURL: connectorClient1._baseUrl,
+        validateStatus: (_) => true
+    });
+    await establishRelationship(connectorClient1, connectorClient2);
+    account2Address = (await connectorClient2.account.getIdentityInfo()).result.address;
+}, getTimeout(30000));
+
+afterAll(() => launcher.stop());
+describe("should log the correlation id", () => {
```
jkoenig134, line 26: newline ...

test/lib/Launcher.ts
```diff
@@ -15,6 +15,7 @@ export type ConnectorClientWithMetadata = ConnectorClient & {
     /* eslint-disable @typescript-eslint/naming-convention */
     _metadata?: Record<string, string>;
     _eventBus?: MockEventBus;
+    _baseUrl?: string;
```
jkoenig134, line 18: why not using launchSimple here?

jkoenig134 (changes requested):

test/correlationId.test.ts
```diff
@@ -0,0 +1,81 @@
...
+let connectorClient2: ConnectorClientWithMetadata;
+let account2Address: string;
+const uuidRegex = new RegExp("[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}");
+
+beforeAll(async () => {
+    [connectorClient1, connectorClient2] = await launcher.launch(2);
+    axiosClient = axios.create({
+        baseURL: connectorClient1._baseUrl,
+        validateStatus: (_) => true
+    });
```
jkoenig134, line 20: if you use the client from connectorClient1 instead you can leave out setting the api key below and you don't have to expose the baseurl:

```suggestion
    axiosClient = connectorClient1["account"]["httpClient"];
```

test/correlationId.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                    "X-API-KEY": "xxx"
+                }
+            }
+        );
+        await connectorClient1._eventBus?.waitForEvent("__headers", (event: any) => {
+            return uuidRegex.test(event.data["x-correlation-id"]);
+        });
+
+        // Found correlation id otherwise waitForEvent would throw an error
+        expect(true).toBe(true);
```
jkoenig134, line 50: this is not how we trick eslint, just add an exclude

test/correlationId.test.ts
```diff
@@ -0,0 +1,81 @@
...
+                    "X-API-KEY": "xxx"
+                }
+            }
+        );
+        await connectorClient1._eventBus?.waitForEvent("__headers", (event: any) => {
+            return event.data["x-correlation-id"] === customCorrelationId;
+        });
+
+        // Found correlation id otherwise waitForEvent would throw an error
+        expect(true).toBe(true);
```
jkoenig134, line 79: see above

test/lib/Launcher.ts
jkoenig134: added a better possiblity in the test

test/correlationId.test.ts
sebbi08: Yes already in work but we need to add the ability in the SDK to add a custom correlation ID.

This will be a bigger change, so it will come in another PR

test/lib/Launcher.ts
sebbi08: I wanted to leave the original event as because I thought the headers would not be there but as it is the webhook they are there anyway.

test/lib/Launcher.ts
```diff
@@ -129,7 +129,8 @@ export class Launcher {
             .use((req, res) => {
                 res.status(200).send("OK");
 
-                eventBus.publish(new DataEvent(req.body.trigger, req.body.data));
+                // eslint-disable-next-line @typescript-eslint/naming-convention
+                eventBus.publish(new DataEvent(req.body.trigger, { ...req.body.data, _headers: req.headers }));
```
jkoenig134, line 133: I didn't meant to write the headers to the event data. Just add a new ConnectorDataEvent that extends DataEvent and add headers.

test/correlationId.test.ts
```diff
@@ -0,0 +1,80 @@
...
+                    expiresAt: DateTime.now().plus({ hour: 1 }).toISO()
+                },
+                peer: account2Address
+            },
+            {
+                headers: {
+                    // eslint-disable-next-line @typescript-eslint/naming-convention
+                    "X-API-KEY": "xxx"
+                }
+            }
```
jkoenig134, line 43: you can remove this

test/correlationId.test.ts
```diff
@@ -0,0 +1,80 @@
...
+                },
+                peer: account2Address
+            },
+            {
+                headers: {
+                    // eslint-disable-next-line @typescript-eslint/naming-convention
+                    "x-correlation-id": customCorrelationId,
+                    // eslint-disable-next-line @typescript-eslint/naming-convention
+                    "X-API-KEY": "xxx"
+                }
```
jkoenig134, line 72: you can remove X-API-KEY and inline the rest

jkoenig134 (approved):

test/correlationId.test.ts
```diff
@@ -0,0 +1,65 @@
...
+            peer: account2Address
+        });
+
+        await connectorClient1._eventBus?.waitForEvent("consumption.outgoingRequestCreated", (event: any) => {
+            return uuidRegex.test(event.headers["x-correlation-id"]);
+        });
+    });
+
+    // eslint-disable-next-line jest/expect-expect
+    test("should log the custom correlation id", async () => {
```
jkoenig134, line 43: not log.

jkoenig134 (changes requested):

jkoenig134 (approved):

PR #224, opened 2024-08-01
Restrict either the SSE or the Sync module to be active but not both

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,8 +37,11 @@ export function createConnectorConfig(overrides?: RuntimeConfig): ConnectorRunti
         connectorConfig.modules.webhooks = _.defaultsDeep(connectorConfig.modules.webhooksV2, connectorConfig.modules.webhooks);
         delete connectorConfig.modules.webhooksV2;
     }
+
     if (connectorConfig.modules.sync.enabled && connectorConfig.modules.sse.enabled) {
-        throw new Error("SSE and Sync cannot be enabled at the same time.");
+        // eslint-disable-next-line no-console
+        console.warn("The SSE and Sync modules cannot be enabled at the same time, the Sync module will be disabled.");
+        connectorConfig.modules.sync.enabled = false;
     }
 
     return connectorConfig;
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -37,6 +37,9 @@ export function createConnectorConfig(overrides?: RuntimeConfig): ConnectorRunti
         connectorConfig.modules.webhooks = _.defaultsDeep(connectorConfig.modules.webhooksV2, connectorConfig.modules.webhooks);
         delete connectorConfig.modules.webhooksV2;
     }
+    if (connectorConfig.modules.sync.enabled && connectorConfig.modules.sse.enabled) {
```
jkoenig134, line 40: newline missing

src/index.ts
```diff
@@ -28,7 +28,7 @@ export function createConnectorConfig(overrides?: RuntimeConfig): ConnectorRunti
         .file("config-env-file", { file: `config/${process.env.NODE_CONFIG_ENV}.json` })
         .file("default-file", { file: "config/default.json" });
 
-    const connectorConfig = nconf.get();
+    const connectorConfig = nconf.get() as ConnectorRuntimeConfig;
```
jkoenig134, line 31: why was the cast not required before?

src/index.ts
```diff
@@ -37,6 +37,9 @@ export function createConnectorConfig(overrides?: RuntimeConfig): ConnectorRunti
         connectorConfig.modules.webhooks = _.defaultsDeep(connectorConfig.modules.webhooksV2, connectorConfig.modules.webhooks);
         delete connectorConfig.modules.webhooksV2;
     }
+    if (connectorConfig.modules.sync.enabled && connectorConfig.modules.sse.enabled) {
+        throw new Error("SSE and Sync cannot be enabled at the same time.");
```
jkoenig134, line 41: wdym by

```suggestion
        console.warn("The SSE and Sync modules cannot be enabled at the same time, the Sync module will be disabled.");
        connectorConfig.modules.sync.enabled = false;
```

jkoenig134: this is not as "harsh" as the other version" as these modules doesn't harm each other.

src/index.ts
sebbi08: Because it actually was any and in theory it is any the whole env is in there, but we use it as `ConnectorRuntimeConfig` anyway later

src/index.ts
sebbi08: Yea way better.

jkoenig134 (approved):

PR #228, opened 2024-08-02
Fix socket hangups during debugging of tests

slavistan:
# Readiness checklist

-   [x] I labeled the PR.

Not reusing sockets gets rid of socket hangups during debugging, when a connection times out while the event loop is frozen.


sebbi08:
@slavistan looks like you used the V5 branch as base and now merge to main

PR #230, opened 2024-08-05
Add code coverage to tests

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.


package.json
jkoenig134, line 1: what's wrong with `jest --coverage`?

jkoenig134: https://github.com/codecov/example-typescript/blob/main/package.json

sebbi08: Not working with child processes https://github.com/jestjs/jest/issues/3190.
So `jest --coverage` would only deliver coverage for the SDK and also only for the build .js files and not the TS once.

Using nyc directly would only show the coverage of the connector, not the SDK.

c8 had the best result in showing both SDK and connector coverage in the TS files.

.github/workflows/test.yml
```diff
@@ -30,6 +30,14 @@ jobs:
           NMSHD_TEST_CLIENTID: test
           NMSHD_TEST_CLIENTSECRET: test
 
+      - name: Code coverage
+        uses: codecov/codecov-action@v4
+        with:
+          fail_ci_if_error: true # optional (default = false)
+          files: ./coverage/cobertura-coverage.xml # optional
+          token: ${{ secrets.CODECOV_TOKEN }} # required
+          verbose: true # optional (default = false)
```
jkoenig134, line 39: comments in pipelines, my nemesis. yeet that pls.

package.json
jkoenig134: so c8 analyzes code that is running in all child processes?? fancy!

package.json
```diff
@@ -76,6 +89,7 @@
         "agentkeepalive": "4.5.0",
         "amqplib": "^0.10.4",
         "axios": "^1.7.2",
+        "c8": "^10.1.2",
```
jkoenig134, line 92: this should be a dev dependency. pls also make sure to re-run `npm i` to bring that to the lockfile.

package.json
```diff
@@ -27,9 +27,10 @@
         "unlink-runtime": "rm -rf node_modules/@nmshd && npm install",
         "tsc": "tsc",
         "tsc:watch": "tsc -w",
-        "test:ci": "npm run build && jest -i",
+        "test:ci": "npm run build && npx c8 jest -i",
```
jkoenig134, line 30: we don't use npx in scripts. npm is resolving executables already internally. same for the other script.

package.json
sebbi08: Normal nyc should do it also, but it somehow does not :D

package.json
```diff
@@ -27,12 +27,12 @@
...
         "test:local": "npm run test:local:mongo",
         "test:local:mongo": "npm run build --workspaces && npm run start:test-env:mongo && DATABASE_CONNECTION_STRING=\"mongodb://localhost:27018\" jest",
-        "test:local:coverage": "npm run build --workspaces && npm run start:test-env:mongo && DATABASE_CONNECTION_STRING=\"mongodb://localhost:27018\" npx c8 jest",
+        "test:local:coverage": "npm run build --workspaces && npm run start:test-env:mongo && DATABASE_CONNECTION_STRING=\"mongodb://localhost:27018\" c8 jest",
         "test:local:ferretdb": "npm run build --workspaces && npm run start:test-env:ferretdb && DATABASE_CONNECTION_STRING=\"mongodb://localhost:27019\" jest",
-        "generate:connectorConfigSchema": "npx ts-json-schema-generator -p ./src/jsonSchemas/connectorConfig.ts -o ./src/jsonSchemas/connectorConfig.json -t \"ConnectorConfig\" --no-top-ref && prettier -w src/jsonSchemas/connectorConfig.json",
+        "generate:connectorConfigSchema": "ts-json-schema-generator -p ./src/jsonSchemas/connectorConfig.ts -o ./src/jsonSchemas/connectorConfig.json -t \"ConnectorConfig\" --no-top-ref && prettier -w src/jsonSchemas/connectorConfig.json",
```
jkoenig134, line 35: we need npx here as this is no dependency.

jkoenig134: and please don't make it a dependency. npx was perfectly fine here.

jkoenig134 (dismissed):

jkoenig134:
oops, approved by accident. The npx stuff has to be fixed.

jkoenig134 (approved):

PR #232, opened 2024-08-06
Add tests for coverage on main branch

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.


sebbi08:
We need to also publish the code coverage from the main branch to get a better comparison

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,38 @@
+name: Coverage
+
+on:
+  push:
+    branches: [main]
```
jkoenig134, line 5: should also be published from release branches

jkoenig134:
is this really necessary? We disabled tests on main a while ago to reduce overall runtime b/c everything is tested before on the branches itself..

.github/workflows/coverage.yml
jkoenig134: ah, right!

sebbi08:
yes, Codecov needs a coverage report for the PR base commit

image

jkoenig134:
> is this really necessary? We disabled tests on main a while ago to reduce overall runtime b/c everything is tested before on the branches itself..

Ok, now I see the Problem. We don't get information about lines changed. https://app.codecov.io/gh/nmshd/connector/pull/233

If this is really necessary I'd say enable the tests for main instead of creating a new workflow here.

jkoenig134 (approved):

PR #241, opened 2024-08-21
Add AutoDecomposeDeletionProposedRelationships Module

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Similar to the AutoAcceptPendingRelationships Module we want to enable the Connector to cleanup Relationships as soon as possible for demos.


config/default.json
```diff
@@ -73,6 +73,11 @@
             "location": "autoAcceptPendingRelationships/AutoAcceptPendingRelationshipsModule",
             "enabled": false
         },
+        "autoDecomposeDeletionProposedRelationships": {
+            "displayName": "Auto Decompose DeletionProposed Pending Relationships",
```
britsta, line 77: Omit the "Pending" in the `"displayName"`: "Auto Decompose DeletionProposed ~~Pending~~ Relationships"

src/modules/autoDecomposeDeletionProposedRelationships/AutoDecomposeDeletionProposedRelationshipsModule.ts
britsta, line 1: It would be nice if we could add a simple test to check that the AutoDecomposeDeletionProposedRelationshipsModule is working properly.

config/default.json
jkoenig134: check!

jkoenig134:
We don't really have a setup to test such modules :/ we can do that l8er when we have it :)

src/modules/autoDecomposeDeletionProposedRelationships/AutoDecomposeDeletionProposedRelationshipsModule.ts
jkoenig134: We don't really have a setup to test such modules :/ we can do that l8er when we have it :)

src/modules/autoDecomposeDeletionProposedRelationships/AutoDecomposeDeletionProposedRelationshipsModule.ts
```diff
@@ -0,0 +1,45 @@
...
+            this.logger.error("Error while decomposing 'DeletionProposed' Relationship:", result.error);
+        }
+    }
+
+    private isIncomingDeletionProposedRelationship(event: RelationshipChangedEvent) {
+        const data = event.data;
+        if (data.status !== RelationshipStatus.DeletionProposed) return false;
+
+        const auditLogEntry = data.auditLog[-1];
+        return auditLogEntry.createdBy !== this.currentIdentity;
```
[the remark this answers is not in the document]
jkoenig134: Oh, that's true! But will we never have a RelationshipChangedEvent with Status DeletionProposed?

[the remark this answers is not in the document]
jkoenig134: We don't :) good catch!

Magnus-Kuhn (approved):

src/modules/autoDecomposeDeletionProposedRelationships/AutoDecomposeDeletionProposedRelationshipsModule.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export default class AutoDecomposeDeletionProposedRelationshipsModule extends ConnectorRuntimeModule<AutoDecomposeDeletionProposedRelationshipsModuleConfiguration> {
+    private currentIdentity: string;
+
+    public init(): void {
+        // Nothing to do here
+    }
+
+    public async start(): Promise<void> {
+        const currentIdentityResult = await this.runtime.getServices().transportServices.account.getIdentityInfo();
```
slavistan, line 14: Should check for error, no?

jkoenig134: no, it will fail automatically when it failed.

slavistan: You mean accessing `.value` in case of an error will raise an exception?

jkoenig134: exactly!

jkoenig134: And as this module is only for internal / testing purposes we don't have to write special error messages. Also I don't see a reason why this call should fail :D

src/modules/autoDecomposeDeletionProposedRelationships/AutoDecomposeDeletionProposedRelationshipsModule.ts
```diff
@@ -0,0 +1,37 @@
...
+        this.currentIdentity = currentIdentityResult.value.address;
+
+        this.subscribeToEvent(RelationshipChangedEvent, this.handleRelationshipChanged.bind(this));
+    }
+
+    private async handleRelationshipChanged(event: RelationshipChangedEvent) {
+        if (event.data.status !== RelationshipStatus.DeletionProposed) return;
+
+        this.logger.info("Incoming 'DeletionProposed' Relationship detected.");
```
slavistan, line 23: I feel like the event is "incoming", not the relationship itself, i.e. `"Incoming 'DeletionProposed' Relationship Event detected."`.

src/ConnectorRuntime.ts
```diff
@@ -36,14 +36,7 @@ export class ConnectorRuntime extends Runtime<ConnectorRuntimeConfig> {
     private accountController: AccountController;
 
     private _transportServices: TransportServices;
-    public get transportServices(): TransportServices {
-        return this._transportServices;
-    }
-
     private _consumptionServices: ConsumptionServices;
```
slavistan, line 39: If you remove the getters I'd suggest giving the private members their proper names back, without the leading underscore.

config/default.json
```diff
@@ -73,6 +73,11 @@
             "location": "autoAcceptPendingRelationships/AutoAcceptPendingRelationshipsModule",
             "enabled": false
         },
+        "autoDecomposeDeletionProposedRelationships": {
```
slavistan, line 76: What do y'all think about shortening the name of this feature to "AutoDecomposeRelationships"? I feel like the rest is redundant.

src/ConnectorRuntime.ts
jkoenig134: our naming sceme is _ for privates, nothing is proper about removing _ here.

config/default.json
jkoenig134: verbosity != redundancy.

src/ConnectorRuntime.ts
slavistan: There are other private members named without any underscores.

jkoenig134: And this invalidates my point?

config/default.json
jkoenig134: Also i just stay in the naming scheme of the Auto..Modules

slavistan (approved):

config/default.json
jkoenig134: And AutoDecomposeRelationships is not precise enough: it's also possible to decompose terminated relationships, so I would exept those Relationships also being decomposed.

PR #246, opened 2024-08-27
Sync Modules does not log errors occuring during the Synchronization

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

This module never logged errors as our usecases does not throw.


tnotheis (approved):

Milena-Czierlinski:
> This module never logged errors as our usecases does not throw.

Is that really the case? What about e.g. when a use case calls a method from a consumption controller and that method throws an error?

jkoenig134:
Then this is wrapped by the usecase. This is the whole idea of usecases..

jkoenig134:
image

PR #248, opened 2024-08-30
Update openapi.yml with correct values for `POST /api/v2/Attributes`

britsta:
# Readiness checklist

-   [x] I labeled the PR.

# Description

When updating the migration guide for the migration from v4 to v5 of the Runtime in the documentation (https://github.com/nmshd/documentation/pull/257), I noticed a few things for the Connector that now have to be adjusted after updating the Runtime version used within the Connector from v4 to v5. 

Has anyone else noticed anything else I could include in this PR?


jkoenig134:
I updated the title to be actually fitting to the changelog. The other fix is only internal and does not really show up for the user.

I'd not stuff more into this PR as it already has to many different changes that are hard to conclude for the changelog.

jkoenig134 (approved):

PR #265, opened 2024-09-12
Cleanup the connector config

sebbi08:
# Readiness checklist

-   [ ] I ensured that the PR title is good enough for the changelog.
-   [x] I labeled the PR.


src/ConnectorRuntimeConfig.ts
```diff
@@ -22,10 +23,6 @@ export interface ConnectorRuntimeConfig extends RuntimeConfig {
     modules: Record<string, ConnectorRuntimeModuleConfiguration>;
 
     infrastructure: {
-        httpServer: {
-            enabled: boolean;
-            apiKey: string;
-            cors?: any;
-        };
+        httpServer: HttpServerConfiguration;
```
jkoenig134, line 26: you have now exposed the port and helmetOptions which we didn't do on purpose. please undo.

sebbi08: That is the ConnectorRuntimeConfig and not the connectorConfig that is used for the config file.

In the schema file, the port is already exposed.

jkoenig134: nvm, I think it's ok.

jkoenig134 (approved):

PR #269, opened 2024-09-23
Remove object loading by id and key

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.


src/modules/coreHttpApi/openapi.yml
```diff
@@ -1909,16 +1909,9 @@ paths:
           application/json:
```
Magnus-Kuhn, line 1909: The description a couple of lines of above still mentions the secret key

jkoenig134: Good catch, I only looked for `secretKey`

Magnus-Kuhn (approved):

Magnus-Kuhn (approved):

PR #285, opened 2024-10-16
Mixed up properties in OpenAPI schemas

britsta:
When loading the newest OpenAPI specification of the Connector into Postman, I've noticed that sometimes the properties of different types are listed together in the body schemas of some API routes, although only one type and thus one set of properties may be used.

# Readiness checklist

-   [x] I labeled the PR.


jkoenig134 (approved):
awesome!

PR #293, opened 2024-10-25
Personalized RelationshipTemplates

Magnus-Kuhn:
# Readiness checklist

-   [X] I added/updated tests.
-   [X] I ensured that the PR title is good enough for the changelog.
-   [X] I labeled the PR.


jkoenig134 (approved):
LGTM, @britsta PTAL

packages/sdk/src/types/relationshipTemplates/requests/GetPeerRelationshipTemplatesRequest.ts
```diff
@@ -3,4 +3,5 @@ export interface GetPeerRelationshipTemplatesRequest {
     expiresAt?: string | string[];
     createdBy?: string | string[];
     maxNumberOfAllocations?: number | number[];
+    forIdentity?: string | string[];
```
britsta, line 6: `string` instead of `string | string[]` should be sufficient as type, because all PeerRelationshipTemplates I have either have `forIdentity` undefined or have specified my own address there, because otherwise I would not have the RelationshipTemplate.

test/relationshipTemplates.test.ts
```diff
@@ -82,6 +82,24 @@ describe("Template Tests", () => {
...
+    });
+
+    test("send and receive a personalized template via token", async () => {
+        const templateToken = await getTemplateToken(client1, (await client2.account.getIdentityInfo()).result.address);
+
+        const response = await client2.relationshipTemplates.loadPeerRelationshipTemplate({
+            reference: templateToken.truncatedReference
+        });
+        expect(response).toBeSuccessful(ValidationSchema.RelationshipTemplate);
+    });
```
britsta, line 102: Can we check as well that an Identity for which the RelationshipTemplate has not been personalized cannot load it?

src/modules/coreHttpApi/openapi.yml
```diff
@@ -3508,6 +3524,10 @@ paths:
                 ephemeral:
                   description: If set to true the token will will not be cached in the database of the connector. Note that you will not be able to fetch this token unless you remember the truncatedReference of the token. Defaults to true. Will be ignored if Accept is set to image/png.
                   type: boolean
+                forIdentity:
+                  $ref: "#/components/schemas/Address"
+                  nullable: true
+                  description: The only Identity that may load this Token. If the RelationshipTemplate can only be loaded by one Identity, that Identity must also be given here.
```
britsta, line 3530: > If the RelationshipTemplate can only be loaded by one Identity

Do you mean, if `maxNumberOfAllocations` of the RelationshipTemplate is set to one? If so, please make that clearer here.

packages/sdk/src/types/relationshipTemplates/requests/GetPeerRelationshipTemplatesRequest.ts
Magnus-Kuhn: That would be an option, the impression of what I have seen in those requests was that we don't think and just allow everything (isOwn has `boolean | boolean[]`). But GetAttributesRequest has almost no arrays? 🤔

src/modules/coreHttpApi/openapi.yml
Magnus-Kuhn: I meant something different, thanks for pointing that out

test/relationshipTemplates.test.ts
Magnus-Kuhn: I only wanted to check that the runtime is called correctly, for that the positive tests are enough. Added checks that the result contains the correct personalization though.

packages/sdk/src/types/relationshipTemplates/requests/GetPeerRelationshipTemplatesRequest.ts
britsta: "Don't think" is a good motto :D For me, both are okay, so throwing away the string array (for the reasons described) or leaving it.

jkoenig134: IMO forIdentity with string-array is valid here, you could match for different identities in one query for example.

britsta: > you could match for different identities in one query for example.

Why it should be possible? All PeerTemplates that I have loaded and can get with this function call have `forIdentity` undefined or with my own address (otherwise, I couldn't load them at the first place).

packages/sdk/src/types/tokens/ConnectorToken.ts
```diff
@@ -5,6 +5,7 @@ export interface ConnectorToken {
     content: unknown;
     createdAt: string;
     expiresAt: string;
+    forIdentity?: string;
```
Milena-Czierlinski, line 8: I guess strictly speaking this and the change in CreateTokenForOwnRelationshipTemplateRequest belong to #294 , but since this will hopefully be finished soon, too, I don't care too much.

jkoenig134: ehm, no. You can have a personalized token for a template so this has to be added now

jkoenig134 (approved):
LGTM

PR #295, opened 2024-10-28
TLS Certificate Pinning and Force TLS Checks

Magnus-Kuhn:
# Readiness checklist

-   [X] I ensured that the PR title is good enough for the changelog.
-   [X] I labeled the PR.

# Description

SDH256 fingerprints can be generated with the following commands:

```sh
echo -n | openssl s_client -connect <hostname>:443 | sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' > cert.pem
openssl x509 -noout -in cert.pem -fingerprint -sha256
```

To test this new feature update `.dev/config.json` with the following entries:

```json
{
    ...
    "pinnedTLSCertificateSHA256Fingerprints": {
        "<hostname>": ["<fingerprint-generated-above>"]
    },
    "enforceCertificatePinning": true
}
```

and start the Connector.


jkoenig134:
I have some overall problems with this PR:

- `pinnedPublicKeys` -> this config key name is not good. YOU know what kind of public keys are meant, but how should anyone else understand this without looking deep in the code how this is used
- why do you extend the transport library config? "pinnedPublicKeys" is no value from the transport lib and should not be placed there

Magnus-Kuhn:
> why do you extend the transport library config?

This was a remnant from when it was set in transport :)

stnmtz (approved):

jkoenig134 (approved):

stnmtz:
Ahh, description no longer fits to implementation ;)

jkoenig134:
@stnmtz which description???? :D

jkoenig134:
@stnmtz & @Magnus-Kuhn I updated the Description to match the new implementation. Should be helpful to update the docs!

PR #299, opened 2024-10-29
Re-add deprecated function deleteThirdPartyOwnedRelationshipAttributeAndNotifyPeer

Milena-Czierlinski:
For backwards compatibility

# Readiness checklist

-   [x] I labeled the PR.

### DIFF
--- a/packages/sdk/src/endpoints/AttributesEndpoint.ts
+++ b/packages/sdk/src/endpoints/AttributesEndpoint.ts
@@ -89,6 +89,15 @@ export class AttributesEndpoint extends Endpoint {
         return await this.delete(`/api/v2/Attributes/ThirdParty/${attributeId}`);
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


jkoenig134:
@Milena-Czierlinski as we are the only consumers we can keep this removed silently! :)

Milena-Czierlinski:
@sebbi08 expressed his concern that the previous release introduced a breaking change for integrators.

jkoenig134:
IMO as the API is unchanged and we don't know about any consumers of the js sdk than us we can ignore it.

jkoenig134:
also I don't expect that someone is already using this functionality

jkoenig134:
@sebbi08 as you requested this, any word from your side for or against this?

sebbi08:
If we are sure that no one uses this, I am also fine with just removing it. Otherwise, this would be a breaking change.

PR #306, opened 2024-11-07
Warn about a weak API-Key at startup

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Additionally it is not possible anymore to start the connector without an API Key in production mode.

### DIFF
--- a/README_dev.md
+++ b/README_dev.md
@@ -143,18 +143,18 @@ npm run test:local -- testSuiteName
     {
       "debug": true,
       "transportLibrary": {
-          "baseUrl": "...",
-          "platformClientId": "...",
-          "platformClientSecret": "..."
+          "baseUrl": "<base-url>",
+          "platformClientId": "<client-id>",
+          "platformClientSecret": "<client-secret>"
       },
       "database": { "driver": "lokijs", "folder": "./" },
       "logging": { "categories": { "default": { "appenders": ["console"] } } },
-      "infrastructure": { "httpServer": { "apiKey": "", "port": 8080 } },
+      "infrastructure": { "httpServer": { "apiKey": "<api-key-or-empty-string>", "port": 8080 } },
       "modules": { "coreHttpApi": { "docs": { "enabled": true } } }
     }
     ```
 
-6. replace ... in the config with real values
+6. replace the placeholders in the config with real values
 7. start the connector using `CUSTOM_CONFIG_LOCATION=./local.config.json node dist/index.js start`
 
 It's now possible to access the connector on port 8080. Validating this is possible by accessing `http://localhost:8080/docs/swagger` in the browser.
--- a/README_dev.md
+++ b/README_dev.md
@@ -149,7 +149,7 @@ npm run test:local -- testSuiteName
       },
       "database": { "driver": "lokijs", "folder": "./" },
       "logging": { "categories": { "default": { "appenders": ["console"] } } },
-      "infrastructure": { "httpServer": { "apiKey": "xxx", "port": 8080 } },
+      "infrastructure": { "httpServer": { "apiKey": "", "port": 8080 } },
       "modules": { "coreHttpApi": { "docs": { "enabled": true } } }
     }
     ```
--- a/src/infrastructure/httpServer/HttpServer.ts
+++ b/src/infrastructure/httpServer/HttpServer.ts
@@ -202,8 +202,9 @@ export class HttpServer extends ConnectorInfrastructure<HttpServerConfiguration>
         const apiKeyPolicy = /^(?=.*[A-Z].*[A-Z])(?=.*[!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~])(?=.*[0-9].*[0-9])(?=.*[a-z].*[a-z]).{30,}$/;
         if (!this.configuration.apiKey.match(apiKeyPolicy)) {
             this.logger.warn(
-                "The configured API key does not meet the requirements. It must be 30 chars and contain at least 1 digit, 1 uppercase, 1 lowercase, and 1 special char."
+                "The configured API key does not meet the requirements. It must be at least 30 characters long and contain at least 2 digits, 2 uppercase letters, 2 lowercase letters and 1 special character (!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~])."
             );
+            this.logger.warn("The API key will be used as is, but it is recommended to change it as it will not be supported in future versions.");
         }
 
         this.app.use(async (req, res, next) => {
--- a/src/infrastructure/httpServer/HttpServer.ts
+++ b/src/infrastructure/httpServer/HttpServer.ts
@@ -1,6 +1,7 @@
 import { sleep } from "@js-soft/ts-utils";
 import { Container } from "@nmshd/typescript-ioc";
 import { Server } from "@nmshd/typescript-rest";
+import { passwordStrength as checkAPIKeyStrength } from "check-password-strength";
 import compression from "compression";
 import correlator from "correlation-id";
 import cors, { CorsOptions } from "cors";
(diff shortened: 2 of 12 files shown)


jkoenig134:
Hmm, some downside - the warning is now spammed in the tests. Maybe it should be disabled when tests are detected.

src/infrastructure/httpServer/HttpServer.ts
```diff
@@ -203,6 +190,40 @@ export class HttpServer extends ConnectorInfrastructure<HttpServerConfiguration>
...
+            }
+        }
+
+        const apiKeyStrength = checkAPIKeyStrength(this.configuration.apiKey, [
+            { id: 0, value: "Very Weak", minDiversity: 0, minLength: 0 },
+            { id: 1, value: "Strong", minDiversity: 0, minLength: 30 }
+        ]);
+        if (apiKeyStrength.value !== "Strong") {
+            this.logger.warn(
+                "The configured API key does not meet the requirements. It must be 30 chars and contain at least 1 digit, 1 uppercase, 1 lowercase, and 1 special char."
```
Milena-Czierlinski, line 209: It must be [at least] 30 chars [long]?

src/infrastructure/httpServer/HttpServer.ts
```diff
@@ -203,6 +190,40 @@ export class HttpServer extends ConnectorInfrastructure<HttpServerConfiguration>
...
+            );
+        }
+
+        this.app.use(async (req, res, next) => {
+            const apiKeyFromHeader = req.headers["x-api-key"];
+            if (!apiKeyFromHeader || apiKeyFromHeader !== this.configuration.apiKey) {
+                // sleep for 1 to 4 seconds
+                await sleep(1000 * (Math.floor(Math.random() * 4) + 1));
```
Milena-Czierlinski, line 217: Why do we do this?

jkoenig134: To make brute-froce way harder.

Milena-Czierlinski:
> the warning is now spammed in the tests

Why don't we have a strong API-Key? 🧐

src/infrastructure/httpServer/HttpServer.ts
jkoenig134: I changed it from "one second" to "frome one to four second" to make it way harder for an attacker to estimate when the timeout hits.

jkoenig134:
@Milena-Czierlinski also an Option, let me think about this.

src/infrastructure/httpServer/HttpServer.ts
```diff
@@ -203,6 +190,40 @@ export class HttpServer extends ConnectorInfrastructure<HttpServerConfiguration>
...
+        if (apiKeyStrength.value !== "Strong") {
+            this.logger.warn(
+                "The configured API key does not meet the requirements. It must be 30 chars and contain at least 1 digit, 1 uppercase, 1 lowercase, and 1 special char."
+            );
+        }
+
+        this.app.use(async (req, res, next) => {
+            const apiKeyFromHeader = req.headers["x-api-key"];
+            if (!apiKeyFromHeader || apiKeyFromHeader !== this.configuration.apiKey) {
+                // sleep for 1 to 4 seconds
```
britsta, line 216: Remove comment please.

jkoenig134: I have added this on purpose b/c I was not sure if the 1-4 seconds logic is understandable without. I'm fine with removing it if everyone understands it.

src/infrastructure/httpServer/HttpServer.ts
```diff
@@ -203,6 +190,40 @@ export class HttpServer extends ConnectorInfrastructure<HttpServerConfiguration>
...
+                    return;
+                case "production":
+                    throw new Error("No API key set in configuration. This is required in production mode.");
+            }
+        }
+
+        const apiKeyStrength = checkAPIKeyStrength(this.configuration.apiKey, [
+            { id: 0, value: "Very Weak", minDiversity: 0, minLength: 0 },
+            { id: 1, value: "Strong", minDiversity: 0, minLength: 30 }
+        ]);
```
britsta, line 206: Do I get it right, that a password of 29 characters and a lot of digitis, special characters and uppercase and lowercase letters is now displayed as "Very Weak"? I get that it does not meet our requirements, but maybe we should call it a "Weak" password then if we don't want to have a finer subdivision of password strengths.

src/infrastructure/httpServer/HttpServer.ts
britsta: I think this line is easily looked-up in case someone is interested in the time of sleep. IMO our guideline to not use comments is more important here.

src/infrastructure/httpServer/HttpServer.ts
jkoenig134: I have to define a "low bound", the name here is irrelevant.

stnmtz:
I'm absolutely no fan of introducing an external password-check library for such a low-level check. I would recommend removing this external dependency for security reasons and match the following regexp which still needs to be fit to our requirements (see https://stackoverflow.com/questions/5142103/regex-to-validate-password-strength).

> ^                         Start anchor
(?=.*[A-Z].*[A-Z])        Ensure string has two uppercase letters.
(?=.*[\.!@#$&*])            Ensure string has one special case letter.
(?=.*[0-9].*[0-9])        Ensure string has two digits.
(?=.*[a-z].*[a-z]) Ensure string has two lowercase letters.
.{8}                      Ensure string is of length 8.
$                         End anchor.

```
passwordPolicy = new RegExp("^(?=.*[A-Z].*[A-Z])(?=.*[!@#$&*])(?=.*[0-9].*[0-9])(?=.*[a-z].*[a-z]).{30,}$")
if (!passwordPolicy.test(password)) {
    console.warn(...)
    // bye bye baby
}
```

jkoenig134:
> I'm absolutely no fan of introducing an external password-check library for such a low-level check

That lib was well maintained and had more features than a simple regex, but I removed it.

I added some more special characters to make the feature way more usable.

README_dev.md
```diff
@@ -149,7 +149,7 @@ npm run test:local -- testSuiteName
       },
       "database": { "driver": "lokijs", "folder": "./" },
       "logging": { "categories": { "default": { "appenders": ["console"] } } },
-      "infrastructure": { "httpServer": { "apiKey": "xxx", "port": 8080 } },
+      "infrastructure": { "httpServer": { "apiKey": "", "port": 8080 } },
```
britsta, line 152: Maybe use `"..."` instead of `""` as placeholder for the `"apiKey"` as it is done for the other fields (`"baseUrl"`, `"platformClientId"`, `"platformClientSecret"`).

jkoenig134: if it was a placeholder I would have used `<...>`

britsta: > It disables the API key.

Good to know!

> if it was a placeholder I would have used `<...>`

I would have appreciated that too (even if the example unfortunately uses `"..."` as a placeholder instead of `"<...>"` so far).

jkoenig134: done.

README_dev.md
britsta, line 1: Should the placeholder for the `apiKey` in the README.md adjusted as well from `"<api-key>"` to `"<api-key-or-empty-string>"`?

jkoenig134: There is no apiKey in README.md?

jkoenig134: ah, you mean for helm. NO, we don't want to have no api key for helm deployments.

britsta: > ah, you mean for helm. NO, we don't want to have no api key for helm deployments.

Yeah, I meant helm :D Alright!

britsta (approved):

stnmtz:
LGTM

stnmtz (approved):

PR #308, opened 2024-11-08
Add identity deletion to the CLI

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.

# Description


jkoenig134:
Oh, I missed `The API is only available when in debug mode.`, but what do we need that for?

sebbi08:
> Oh, I missed `The API is only available when in debug mode.`, but what do we need that for?

That was a request from the APP to make the automated tests easier

sebbi08:
> Also, we need the connector to refuse startup if the identity is in deletion.

Yea will check that

jkoenig134:
> > Oh, I missed `The API is only available when in debug mode.`, but what do we need that for?
> 
> That was a request from the APP to make the automated tests easier

I really don't see why this is stuffed into this PR.

src/ConnectorRuntime.ts
```diff
@@ -42,6 +42,8 @@ export class ConnectorRuntime extends Runtime<ConnectorRuntimeConfig> {
         return this.runtimeConfig.debug ? "debug" : "production";
     }
 
+    public isDebugMode = (): boolean => this.runtimeConfig.debug;
```
jkoenig134, line 45: why do we have the debug mode? please use this instead.

sebbi08: I even put it directly below it and didn't notice :D

sebbi08:
> > > Oh, I missed `The API is only available when in debug mode.`, but what do we need that for?
> > 
> > 
> > That was a request from the APP to make the automated tests easier
> 
> I really don't see why this is stuffed into this PR.

Because it both is part of the identity deletion feature of the connector, also it is part of the same ticket.

src/modules/coreHttpApi/controllers/IdentityDeletionController.ts
```diff
@@ -0,0 +1,24 @@
...
+export class IdentityDeletionController extends BaseController {
+    public constructor(@Inject private readonly transportServices: TransportServices) {
+        super();
+    }
+
+    @DELETE
+    @Path("/")
+    @Accept("application/json")
+    public async getIdentityInfo(@Context context: ServiceContext): Promise<Envelope> {
+        if (!context.request._debugMode) {
```
jkoenig134, line 18: I'd rather like this controller to not be loaded at all instead of passing a debug mode flag that can easily get lost and then show this route to the public.

As said before, please remove all the routes and put them to another PR, we need to find better ways to provide debug apis - as this is not the way.

src/cli/commands/identity/status.ts
```diff
@@ -0,0 +1,40 @@
...
+export class IdentityStatus extends BaseCommand {
+    protected async runInternal(): Promise<void> {
+        await this.createRuntime();
+        if (!this.cliRuntime) {
+            throw new Error("Failed to inizitialize runtime");
+        }
+
+        try {
+            const identityInfoResult = await this.cliRuntime.getServices().transportServices.account.getIdentityInfo();
+            const identityDeletionProcessesResult = await this.cliRuntime.getServices().transportServices.identityDeletionProcesses.getActiveIdentityDeletionProcess();
```
britsta, line 24: Please change variable name to use singular instead of plural (there can be at most one active IdentityDeletionProcess as far as I know).

src/cli/commands/identity/status.ts
```diff
@@ -0,0 +1,40 @@
...
+
+            const identityInfo = identityInfoResult.value;
+            let message = `Id: ${identityInfo.address}`;
+
+            if (identityDeletionProcessesResult.isSuccess) {
+                const identityDeletionProcesses = identityDeletionProcessesResult.value;
```
britsta, line 30: Please change variable name to use singular instead of plural (there can be at most one active IdentityDeletionProcess as far as I know).

src/cli/commands/identity/status.ts
```diff
@@ -0,0 +1,40 @@
...
+                message += `\nIdentity deletion status: ${identityDeletionProcesses.status}`;
+                message += `\nEnd of approval period: ${DateTime.fromISO(identityDeletionProcesses.approvalPeriodEndsAt ?? "").toLocaleString()}`;
```
britsta, line 32: If `approvalPeriodEndsAt` is undefined, I don't like the output `End of approval period: ""`. Please find a better solution.

src/cli/commands/identity/status.ts
```diff
@@ -0,0 +1,40 @@
...
+                message += `\nEnd of grace period: ${DateTime.fromISO(identityDeletionProcesses.gracePeriodEndsAt ?? "").toLocaleString()}`;
```
britsta, line 33: If `gracePeriodEndsAt` is undefined, I don't like the output `End of grace period: ""`. Please find a better solution.

src/cli/commands/identity/status.ts
britsta, line 32: If the IdentityDeletionProcess is already approved, this information isn't needed.

src/cli/commands/identity/status.ts
```diff
@@ -0,0 +1,40 @@
...
+    await new IdentityStatus().run(config);
+};
+export const yargsIdentityStatusCommand: CommandModule<{}, ConfigFileOptions> = {
+    command: "status",
+    describe: "show the status of the identity",
+    handler: identityStatusHandler,
+    builder: configOptionBuilder
+};
+
+export class IdentityStatus extends BaseCommand {
```
britsta, line 15: Should IdentityStatus only refer to the IdentityDeletionProcessStatus? If so, I would expect another name.

test/modules/cli/identityDeletion/identityDeletion.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        const consoleSpy = jest.spyOn(console, "log");
+        await identityDeletionInitHandler({});
+        expect(consoleSpy).toHaveBeenCalledWith("Identity deletion initiated");
+        expect(consoleSpy).toHaveBeenCalledTimes(1);
+    });
+
+    test("cancel identity deletion", async () => {
+        const consoleSpy = jest.spyOn(console, "log");
+        await identityDeletionCancelHandler({});
+        expect(consoleSpy).toHaveBeenCalledWith("Identity deletion canceled");
```
britsta, line 27: Write "cancelled" instead of "canceled" please.

src/cli/commands/identityDeletion/cancel.ts
```diff
@@ -0,0 +1,29 @@
...
+    protected async runInternal(): Promise<void> {
+        await this.createRuntime();
+        if (!this.cliRuntime) {
+            throw new Error("Failed to initialize runtime");
+        }
+
+        const identityDeletionCancelationResult = await this.cliRuntime.getServices().transportServices.identityDeletionProcesses.cancelIdentityDeletionProcess();
+
+        if (identityDeletionCancelationResult.isSuccess) {
+            this.log.log("Identity deletion canceled");
```
britsta, line 24: Write "cancelled" instead of "canceled" please.

src/cli/commands/identity/status.ts
sebbi08: Na this will show some more info like the identity string and if there also show the identity deletion

PR #309, opened 2024-11-11
Some objects in the OpenAPI spec are incorrect

jkoenig134:
# Readiness checklist

-   [x] I labeled the PR.

# Description

Fixes nmshd/feedback#29

Fixed issues:
- `@type` is displayed as required in `RequestResponseContent`
- `peerIdentity` is missing in `Relationship`

### DIFF
--- a/src/modules/coreHttpApi/openapi.yml
+++ b/src/modules/coreHttpApi/openapi.yml
@@ -5614,6 +5614,20 @@ components:
             - $ref: "#/components/schemas/Address"
           nullable: false
           description: The address of the peer identity.
+        peerIdentity:
+          type: object
+          additionalProperties: false
+          properties:
+            address:
+              allOf:
+                - $ref: "#/components/schemas/Address"
+              nullable: false
+            publicKey:
+              type: string
+              nullable: false
+          required:
+            - address
+            - publicKey
         peerDeletionInfo:
           type: object
           additionalProperties: false
--- a/src/modules/coreHttpApi/openapi.yml
+++ b/src/modules/coreHttpApi/openapi.yml
@@ -5113,7 +5113,6 @@ components:
               - $ref: "#/components/schemas/RequestResponseContentItem"
               - $ref: "#/components/schemas/RequestResponseContentItemGroup"
       required:
-        - "@type"
         - result
         - requestId
         - items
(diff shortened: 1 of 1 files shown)


jkoenig134:
I don't get your question. In a perfect world openapi and the results would be 1:1 the same. It's hard to archieve that, so I fixed these bugs.
If you find more feel free to mention that :)

src/modules/coreHttpApi/openapi.yml
```diff
@@ -5615,6 +5614,20 @@ components:
             - $ref: "#/components/schemas/Address"
           nullable: false
           description: The address of the peer identity.
+        peerIdentity:
```
Milena-Czierlinski, line 5617: `peerIdentity` must be added below to required. Also, to be more consistent with our code, I'd prefer it if you move this below `peerDeletionInfo`.

jkoenig134: It's consistent to how the object will occur so I leave it where it is.

Milena-Czierlinski:
I had trouble finding the data structure that is returned as a result so that I might compare them. Where do I see how the returned object will occur?

Another question that is principally unrelated to your changes: Why do we have `peer` and `peerIdentity`?

jkoenig134:
> I had trouble finding the data structure that is returned as a result so that I might compare them. Where do I see how the returned object will occur?
> 
> Another question that is principally unrelated to your changes: Why do we have `peer` and `peerIdentity`?

The Data structure is RelationshipDTO.

Regarding `peer` and `peerIdentity`: I think for easier / more performant querying.

Milena-Czierlinski:
And is `@type` never required?

jkoenig134:
`@type` is not rendered out by ts-serval when the property can only have one `@type`, to save bandwidth.

britsta:
> Regarding `peer` and `peerIdentity`: I think for easier / more performant querying.

Should we also add the `peerIdentity` property to the documentation of the Relationship in the Data Model Overview then? I also thought it was weird to have a `peer` property in addition to the `peerIdentity` property. I wonder how we can present this redundance in a meaningful way in the documentation.

jkoenig134:
we should document it, yes.

britsta (approved):

PR #310, opened 2024-11-11
Add identity deletion debug APIs

sebbi08:
# Readiness checklist

-   [ ] I ensured that the PR title is good enough for the changelog.
-   [x] I labeled the PR.


src/modules/coreHttpApi/debug-controllers/IdentityDeletionController.ts
```diff
@@ -0,0 +1,20 @@
...
+import { Envelope } from "../../../infrastructure";
+import { BaseController } from "../common/BaseController";
+
+@Path("/api/v2/IdentityDeletion")
+export class IdentityDeletionController extends BaseController {
+    public constructor(@Inject private readonly transportServices: TransportServices) {
+        super();
+    }
+
+    @DELETE
```
jkoenig134, line 13: I see what you did here - "deleting" the identity using the delete keyword, but this is not REST conform.

1. rename everything here from IdentityDeletion to IdentityDeletionProcesses
2. make this a POST
3. rename this function please, and be more cautious with copy pasting code

Additionally this feature is non-sense if we don't allow to cancel / query the identity deletion process, please add this as routes.

src/modules/coreHttpApi/debug-controllers/IdentityDeletionController.ts
jkoenig134, line 1: this is not tested.

sebbi08: should we test the debug API? 

Also do you want it to be added to the SDK or should we tests it without the SDK?

jkoenig134: yes, test it, but only with axios

jkoenig134 (approved):

PR #323, opened 2024-12-06
Add an option to add a correlation ID to a request

sebbi08:
# Readiness checklist

-   [x] I labeled the PR.
-   [x] I self-reviewed the PR.


sebbi08:
@jkoenig134 what do you think about this ?

Another option would be to tie the client to a correlation ID and "force" the user to use multiple clients.

I am also open for opinions if we should expose those utility functions "startCorrelation", "createCorrelationId" or if we only should document this.

jkoenig134:
I don't really like the idea to add the correlation-id package to the sdk. We should instead provide a possibility to add a correlation id for every request (send the header `X-CORRELATION-ID`)

PR #343, opened 2025-01-10
Add the `identity init` command to the Connector CLI

sebbi08:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [ ] I self-reviewed the PR.

### DIFF
--- a/src/cli/BaseCommand.ts
+++ b/src/cli/BaseCommand.ts
@@ -1,7 +1,14 @@
+import { IDatabaseConnection } from "@js-soft/docdb-access-abstractions";
+import { LokiJsConnection } from "@js-soft/docdb-access-loki";
+import { MongoDbConnection } from "@js-soft/docdb-access-mongo";
+import { NodeLoggerFactory } from "@js-soft/node-logger";
+import { EventEmitter2EventBus } from "@js-soft/ts-utils";
+import { Transport } from "@nmshd/transport";
 import yargs from "yargs";
 import { ConnectorRuntime } from "../ConnectorRuntime";
 import { ConnectorRuntimeConfig } from "../ConnectorRuntimeConfig";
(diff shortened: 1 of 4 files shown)


src/cli/BaseCommand.ts
```diff
@@ -46,7 +55,50 @@ export abstract class BaseCommand {
...
+
+        const mongodbConnection = new MongoDbConnection(runtimeConfig.database.connectionString);
+
+        try {
+            await mongodbConnection.connect();
+        } catch (e) {
+            throw new Error(`Could not connect to the configured database. Try to check the connection string and the database status. Root error: ${e}`);
         }
+        return mongodbConnection;
     }
```
sebbi08, line 102: @jkoenig134 I currently don't like this as it duplicates the DB creation code from the ConnectorRuntime.
Best would be to export the DB creation out of the ConnectorRuntime, but don't know where it fits best

jkoenig134: WTB just starting the runtime with `allowIdentityCreation` set to true in this command? Then the runtime does the rest.

sebbi08: Yea, but we then could not distinguish if the identity already existed or if it was just created. Don't know if we need that, but yea, that would make it way easier.

jkoenig134: I'll think about it. I like the idea to throw an error if the identity already exists..

sebbi08: I now build it so that the already thrown exception is used.

I don't think that throwing an exception when the identity already exists is a good idea because we would need an exception for our test, or rewrite all that "reuse" an existing identity

jkoenig134:
I put this back to draft as it will not land in v7

PR #350, opened 2025-01-17
Custom Connector Module Support

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

Add support for custom connector modules via a npm peer dependency.


.dockerignore
```diff
@@ -1,4 +1,4 @@
-packages
+packages/sdk
```
tnotheis, line 1: Maybe it would be better to still ignore the whole "packages" folder, and only unignore the "connector-types" package.

packages/connector-types/src/infrastructure/httpServer/HttpMethod.ts
tnotheis, line 1: Since we don't know which endpoints someone "external" might want to add, we should probably allow all HTTP methods

packages/connector-types/src/infrastructure/httpServer/IHttpServer.ts
```diff
@@ -0,0 +1,11 @@
+import { RequestHandler } from "express";
+import { ConnectorInfrastructure } from "../ConnectorInfrastructure";
+import { HttpMethod } from "./HttpMethod";
+
+export interface IHttpServer extends ConnectorInfrastructure {
+    start(): Promise<void>;
+    stop(): void;
```
tnotheis, line 7: I don't know if it's a good idea to expose these two methods on the same interface as the methods for configuring the http server.
But on the other hand, for now you probably only install modules that you develop yourself. So if you screw up and call `stop` accidentally, it's your own fault.

tnotheis (changes requested):

.dockerignore
jkoenig134: true.

tnotheis (approved):

tnotheis (dismissed):

tnotheis (approved):

PR #456, opened 2025-05-14
Switch the default QR code format to the new one and remove the old one completely

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.

# Description

This is breaking as integrators have to actively change the behavior if they want to use the old reference format in QR codes.

### DIFF
--- a/src/modules/coreHttpApi/openapi.yml
+++ b/src/modules/coreHttpApi/openapi.yml
@@ -2131,7 +2131,7 @@ paths:
               - $ref: "#/components/schemas/FileReferenceTruncated"
         - in: query
           name: oldQRCodeFormat
-          description: If set to true, the QRCode will contain the old url format. This is only relevant if the accept header is set to `image/png`.
+          description: If set to true, the QRCode will contain the old url format. This is only relevant if the accept header is set to `image/png`. This option is deprecated and will be removed in the future.
           required: false
           schema:
             type: boolean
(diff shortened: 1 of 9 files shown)


jkoenig134:
An alternative would be to completely remove the old format, but I don't know it this would be too breaking..

Milena-Czierlinski:
> An alternative would be to completely remove the old format, but I don't know it this would be too breaking..

Is there any reason for them to use the old format?

src/modules/coreHttpApi/openapi.yml
```diff
@@ -2130,8 +2130,8 @@ paths:
               - $ref: "#/components/schemas/FileID"
               - $ref: "#/components/schemas/FileReferenceTruncated"
         - in: query
-          name: newQRCodeFormat
-          description: If set to true, the QRCode will contain the new url format. This is only relevant if the accept header is set to `image/png`.
+          name: oldQRCodeFormat
+          description: If set to true, the QRCode will contain the old url format. This is only relevant if the accept header is set to `image/png`. This option is available to provide a grace period for all apps to support the new QR code format and will be removed in the future.
```
britsta, line 2134: Just as you write `"QR code"` in your added sentence, please also change `"QRCode"` to `"QR code"` in the first sentence.

The same applies to the other descriptions.

```suggestion
          description: If set to true, the QR code will contain the old url format. This is only relevant if the accept header is set to `image/png`. This option is available to provide a grace period for all apps to support the new QR code format and will be removed in the future.
```

jkoenig134: As I removed that backwards compatibility this is now outdated :)

PR #458, opened 2025-05-14
Simplify QR code generation

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


packages/types/src/infrastructure/httpServer/BaseController.ts
```diff
@@ -62,19 +62,15 @@ export abstract class BaseController {
...
     }
 
-    protected async qrCode<T extends Result<any>>(
-        result: T,
-        qrPredicate: (result: T) => Promise<QRCode>,
-        filename: string,
-        response: express.Response,
-        status: number
-    ): Promise<void> {
+    protected async qrCode<T extends Result<{ reference: { url: string } }>>(result: T, filename: string, response: express.Response, status: number): Promise<void> {
```
tnotheis, line 65: You might as well only pass the url.

```suggestion
    protected async qrCode(url: string, filename: string, response: express.Response, status: number): Promise<void> {
```

tnotheis (approved):

packages/types/src/infrastructure/httpServer/BaseController.ts
jkoenig134: As written per Teams, not possible as this function should call guard on the Result.

PR #487, opened 2025-05-23
Enforce uniqueness of arrays and add support for `wasAutomaticallyDecided` in the request

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


src/modules/coreHttpApi/controllers/IncomingRequestsController.ts
```diff
@@ -21,6 +22,10 @@ export class IncomingRequestsController extends BaseController {
     @Path("/:id/Accept")
     @Accept("application/json")
     public async accept(@PathParam("id") requestId: string, request: any): Promise<Envelope> {
+        if (request?.decidedByAutomation === true) {
+            throw new ApplicationError("error.connector.incomingRequests.decidedByAutomation", "Decided by automation is not allowed for this endpoint.");
+        }
+
```
Milena-Czierlinski, line 28: We should probably do the same for CanAccept and CanReject.

jkoenig134: those routes completely ignore that param, so I don't see a reason to do that.

jkoenig134: The thing is, noone should and will add this decidedByAutomation param.

jkoenig134: But IF, it's only a problem in the two routes I added the validation

jkoenig134: changed.

Milena-Czierlinski (approved):

packages/sdk/src/types/requests/ConnectorRequest.ts
```diff
@@ -19,6 +19,7 @@ export interface ConnectorRequest {
             reference: string;
         };
     };
+    wasAutomaticallyDecided?: true;
```
britsta, line 22: The property `wasAutomaticallyDecided` of the Request documented in the `openapi.yml` is missing.

Also I wonder if we should add this property to the query incoming Requests use case in the Runtime and to the corresponding Connector route as well.

jkoenig134: 3 hours after the PR was opened and 8 minutes after it was merged :(

jkoenig134: > Also I wonder if we should add this property to the query incoming Requests use case in the Runtime and to the corresponding Connector route as well.

yep :)

britsta: > 3 hours after the PR was opened and 8 minutes after it was merged :(

Yeah :( Was reviewing it and then it has been merged before I could send it.

PR #507, opened 2025-06-04
Allow loading items using url references

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


src/modules/coreHttpApi/openapi.yml
```diff
@@ -4536,9 +4537,9 @@ paths:
                   type: string
                   format: byte
                   nullable: false
-                  description: The base64 encoded truncated reference of the Token, which actually consists of all information to get and decrypt it.
+                  description: The base64 encoded truncated reference or url reference of the Token, which actually consists of all information to get and decrypt it.
                 ephemeral:
-                  description: If set to true the token will will not be cached in the database of the Connector. Note that you will not be able to fetch this token unless you remember the truncatedReference of the token. Defaults to false.
+                  description: If set to true the token will will not be cached in the database of the Connector. Note that you will not be able to fetch this token unless you remember the reference of the token. Defaults to false.
```
britsta, line 4542: Typo that can be resolved if you want. Same for other adjusted sentences within this PR (`"will will"` instead of `"will"`).

```suggestion
                  description: If set to true the token will not be cached in the database of the Connector. Note that you will not be able to fetch this token unless you remember the reference of the token. Defaults to false.
```

britsta (approved):

PR #613, opened 2025-08-12
Allow multiple bundled config files from multiple custom modules

sebbi08:
# Readiness checklist

- [ ] I added/updated tests.
- [x] I labeled the PR.
- [x] I self-reviewed the PR.


jkoenig134 (changes requested):
completely destroyed lockfile. Not thought about conflicting configs and config order.

jkoenig134:
@Milena-Czierlinski & @britsta what's your thoughts about this? I have serious concerns about merging this..

sebbi08:
> TBH, I'm not a big fan of this. How do you make sure that:
> 
> 1. bundled configs from different modules doesn't conflict
> 2. you have those configs in the correct order
> 
> I think one config file must be enough.

Yea I see those issues but how should a module define it's own location without using a bundled.config.json when using multiple modules

PR #656, opened 2025-09-10
The reference can not be parsed by old runtime versions because of tailing '&'

sebbi08:
# Readiness checklist

- [ ] I added/updated tests.
- [ ] I ensured that the PR title is good enough for the changelog.
- [ ] I labeled the PR.
- [ ] I self-reviewed the PR.


jkoenig134:
should be against main. I'll do that.

PR #693, opened 2025-10-06
SSE is not reconnecting after a 5xx HTTP error code

jkoenig134:
# Readiness checklist

- [ ] I added/updated tests.
- [ ] I ensured that the PR title is good enough for the changelog.
- [ ] I labeled the PR.
- [ ] I self-reviewed the PR.

# Description

As described in https://stackoverflow.com/a/54385402/10029378 EventSources completely drop the connection if a 502 occurred. As restarts in the backbone _can_ cause 5xx HTTP errors we have to properly check the EventSource if it's still running and if not do a manual reconnect.

Additionally this PR introduces a fallback for the time SSE is not available and syncs every 3 seconds.

### DIFF
--- a/src/modules/sse/SseModule.ts
+++ b/src/modules/sse/SseModule.ts
@@ -27,7 +27,7 @@ export class SseModule extends ConnectorRuntimeModule<SseModuleConfiguration> {
             try {
                 this.eventSource.close();
             } catch (error) {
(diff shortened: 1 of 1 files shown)


jkoenig134:
> LGTM, but maybe wait also for @tnotheis review.

... YOLO!

